PR #2, opened 2020-06-16
Connect azure-c-pal

parth21999:


mattdurak:
> #define CLDS_ATOMICS_H

I guess a future change would remove this and switch to the interlocked module we have now instead.

---
Refers to: inc/clds/clds_atomics.h:4 in 6725f6f. [](commit_id = 6725f6f62b1af47303d999d1cc9ec8fdc6baec94, deletion_comment = False)

PR #5, opened 2020-07-31
propagate gballoc hl

anporumb:


tests/clds_hash_table_int/clds_hash_table_int.c
```diff
@@ -12,13 +12,19 @@
...
-#include "azure_macro_utils/macro_utils.h"
 #include "testrunnerswitcher.h"
-#include "azure_c_pal/gballoc.h"
+
+#include "azure_macro_utils/macro_utils.h"
+
+#include "azure_c_logging/xlogging.h"
+
+#include "azure_c_pal/gballoc_hl.h"
+#include "azure_c_pal/gballoc_hl_redirect.h"
```
mattdurak, line 22: >gballoc_hl_redirect [](start = 22, length = 19)

I thought we didn't need this in int tests? And don't  we need gballoc_hl_init calls?

tests/clds_singly_linked_list_int/clds_singly_linked_list_int.c
```diff
@@ -11,15 +11,20 @@
...
 #include "azure_macro_utils/macro_utils.h"
+
 #include "testrunnerswitcher.h"
 
-#include "windows.h"
+#include "azure_c_logging/xlogging.h"
+
 #include "azure_c_pal/timer.h"
-#include "azure_c_pal/gballoc.h"
+#include "azure_c_pal/gballoc_hl.h"
```
mattdurak, line 23: >gballoc_hl [](start = 22, length = 10)

gballoc_hl_init?

mattdurak: (here and everywhere int test at least)

---
In reply to: [463739376](https://github.com/Azure/clds/pull/5#discussion_r463739376) [](ancestors = 463739376)

tests/clds_hash_table_int/clds_hash_table_int.c
anporumb: Added gballoc_hl_init calls here and everywhere.

About "not needing it" - answer is "maaaaaybe we don't". Here's why

Assume test code uses gballoc_hl to allocate some memory. That memory is then passed to code under test. Now... the code under test ... if it want to free the said memory... NEEDS to calls the same gballoc_hl_free. Or the opposide: test code allocates some memory with good old malloc. This memory is passed to code under test. While not being nice, code under test wants to free it. Then code under test NEEDS to use good old free. 

When ownership of memory is not transferred (and that is the case with mostly all of our tests), it is fine to use good old malloc in test code and code under test can use gballoc_hl_malloc.

But to be safe, I would now (cough) say : use the same everywhere. Same = gballoc_hl. 

This sort of ripples all the way down back into c_pal :(, Simply because I forgot about it. Oh well - those who check in and our have always something to check in.

---
In reply to: [463739137](https://github.com/Azure/clds/pull/5#discussion_r463739137) [](ancestors = 463739137)

tests/clds_singly_linked_list_int/clds_singly_linked_list_int.c
anporumb: Yeah, also unittests need it (they need the real one). If the unittest hooks to gballoc_hl_malloc - it will end up calling gballoc_hl_malloc. Now that can return some hard coded value... or can call into real_gballoc_hl_malloc.

real_gballoc_malloc should have received a real_gballoc_hl_init before it got called.

Now it so happens that these are "passthrough" and that means that they don't actually need an init and that was the reason why tests were passing. Suspicion that when win32heap they would fail :)

---
In reply to: [463739566](https://github.com/Azure/clds/pull/5#discussion_r463739566) [](ancestors = 463739566,463739376)

PR #7, opened 2020-08-09
Fix extraneous skipped seq no

dcristoloveanu:
- Improve hash table int tests to check that all sequence numbers are actually indicated
- Improve sorted list int tests to check that all sequence numbers are actually indicated
- Fix extraneous skipped seq no in insert/delete for sorted list

### DIFF
--- a/tests/clds_hash_table_int/clds_hash_table_int.c
+++ b/tests/clds_hash_table_int/clds_hash_table_int.c
@@ -16,6 +16,8 @@
 
 #include "azure_macro_utils/macro_utils.h"
 
+#include "azure_c_pal/interlocked_hl.h"
+
 #include "azure_c_logging/xlogging.h"
 
 #include "azure_c_pal/gballoc_hl.h"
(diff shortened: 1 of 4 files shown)


tests/clds_hash_table_int/clds_hash_table_int.c
```diff
@@ -1637,24 +1658,51 @@ static int chaos_thread(void* arg)
...
     CHAOS_TEST_CONTEXT* chaos_test_context = (CHAOS_TEST_CONTEXT*)malloc(sizeof(CHAOS_TEST_CONTEXT) + (sizeof(CHAOS_TEST_ITEM_DATA) * CHAOS_ITEM_COUNT));
     ASSERT_IS_NOT_NULL(chaos_test_context);
 
+    for (i = 0; i < TEST_SEQ_NO_QUEUE_SIZE; i++)
+    {
+        (void)InterlockedExchange(&chaos_test_context->seq_numbers[i], SEQ_NO_NOT_USED);
+    }
+
     (void)InterlockedExchange64(&chaos_test_context->seq_no_count, 0);
+    (void)InterlockedExchange64(&chaos_test_context->next_seq_no_to_ack, 1);
```
mattdurak, line 1698: >next_seq_no_to_ack [](start = 53, length = 18)

This isn't volatile and isn't using interlocked in the other thread. Should it just be a local?

mattdurak: Was also wondering if something like this should check that we don't go all the way around the queue and overwrite things (probably unlikely with the size of the queue though...)

---
In reply to: [468054641](https://github.com/Azure/clds/pull/7#discussion_r468054641) [](ancestors = 468054641)

tests/clds_sorted_list_int/clds_sorted_list_int.c
```diff
@@ -1497,22 +1545,51 @@ static int chaos_thread(void* arg)
...
     CHAOS_TEST_CONTEXT* chaos_test_context = (CHAOS_TEST_CONTEXT*)malloc(sizeof(CHAOS_TEST_CONTEXT) + (sizeof(CHAOS_TEST_ITEM_DATA) * CHAOS_ITEM_COUNT));
     ASSERT_IS_NOT_NULL(chaos_test_context);
 
+    for (i = 0; i < TEST_SEQ_NO_QUEUE_SIZE; i++)
+    {
+        (void)InterlockedExchange(&chaos_test_context->seq_numbers[i], SEQ_NO_NOT_USED);
+    }
+
+    (void)InterlockedExchange64(&chaos_test_context->seq_no_count, 0);
+    (void)InterlockedExchange64(&chaos_test_context->next_seq_no_to_ack, 1);
```
mattdurak, line 1585: >InterlockedExchange64 [](start = 10, length = 21)

same comment about not using interlocked/volatile

tests/clds_hash_table_int/clds_hash_table_int.c
dcristoloveanu: Fixed.

---
In reply to: [468057928](https://github.com/Azure/clds/pull/7#discussion_r468057928) [](ancestors = 468057928,468054641)

tests/clds_sorted_list_int/clds_sorted_list_int.c
dcristoloveanu: Fixed.

---
In reply to: [468058351](https://github.com/Azure/clds/pull/7#discussion_r468058351) [](ancestors = 468058351)

tests/clds_hash_table_int/clds_hash_table_int.c
dcristoloveanu: For the second part as discussed in person, the count of indicated sequence numbers is also checked. Thus it is not really possible to skip a sequence number, wrap around and thus mask a bug.

---
In reply to: [468101203](https://github.com/Azure/clds/pull/7#discussion_r468101203) [](ancestors = 468101203,468057928,468054641)

PR #17, opened 2020-10-26
Make all unit tests run on Linux also

dcristoloveanu:
Make all unit tests run on Linux also


CMakeLists.txt
```diff
@@ -152,32 +152,32 @@ endif()
...
 endif()
 
 set(clds_c_files
+    ./src/clds_hazard_pointers.c
+    ./src/clds_sorted_list.c
+    ./src/clds_st_hash_set.c
     ./src/lock_free_set.c
 )
 
 if (WIN32)
```
mattdurak, line 176: so we still have 2 modules that don't work on Linux?

mattdurak (approved):
:shipit:

CMakeLists.txt
dcristoloveanu: Cheers for pointing it out. No, all files build now on Linux, so no more if WIN32 here

---
In reply to: [512049942](https://github.com/Azure/clds/pull/17#discussion_r512049942) [](ancestors = 512049942)

PR #22, opened 2020-10-29
Enable linux for all tests

dcristoloveanu:


CMakeLists.txt
```diff
@@ -31,6 +31,10 @@ set(run_int_tests OFF)
 set(run_traceability OFF)
 set(run_perf_tests OFF)
 
+if(WIN32)
+else()
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g")
```
mattdurak, line 36: >${CMAKE_C_FLAGS} -g" [](start = 19, length = 20)

Still wondering if we can just update the cmake command in build.sh instead with `-DCMAKE_BUILD_TYPE=Debug` or `-DCMAKE_BUILD_TYPE=RelWithDebInfo` ...

tests/clds_hash_table_int/clds_hash_table_int.c
```diff
@@ -487,7 +487,14 @@ typedef struct CHAOS_THREAD_DATA_TAG
 
 #define CHAOS_THREAD_COUNT  16
 #define CHAOS_ITEM_COUNT    10000
-#define CHAOS_TEST_RUNTIME  30000
+
+#ifdef _MSC_VER
+// on Windows run for longer. Normally this should be passed as an argument, but this should also do for now.
+#define CHAOS_TEST_RUNTIME  300000
+#else
+// setting this way lower as we run with Helgrind on Linux and that is ... slow
```
mattdurak, line 495: >as we run with Helgrind on Linux and that is ... slow [](start = 26, length = 53)

I assume we don't have a way to detect we are using Helgrind and make this run longer if not (since we run the Linux tests normally and with Helgrind/valgrind/etc)

tests/clds_sorted_list_int/clds_sorted_list_int.c
```diff
@@ -164,7 +164,7 @@ typedef struct CHAOS_THREAD_DATA_TAG
 
 #define CHAOS_THREAD_COUNT  16
 #define CHAOS_ITEM_COUNT    10000
-#define CHAOS_TEST_RUNTIME  300000
+#define CHAOS_TEST_RUNTIME  30000
```
mattdurak, line 167: >30000 [](start = 28, length = 5)

Should this be reduced on Linux only? Also seeing this I'm wondering what the units are...

mattdurak (approved):
:shipit:

CMakeLists.txt
dcristoloveanu: I'll have another pass for that. Sorry, too many things piling up with these changes for Linux

---
In reply to: [518861336](https://github.com/Azure/clds/pull/22#discussion_r518861336) [](ancestors = 518861336)

tests/clds_hash_table_int/clds_hash_table_int.c
dcristoloveanu: Hmm, we actually do, we have a define for usage of valgrind now. I'll switch to using that.

---
In reply to: [518862947](https://github.com/Azure/clds/pull/22#discussion_r518862947) [](ancestors = 518862947)

tests/clds_sorted_list_int/clds_sorted_list_int.c
dcristoloveanu: Fixed.

---
In reply to: [518863975](https://github.com/Azure/clds/pull/22#discussion_r518863975) [](ancestors = 518863975)

PR #25, opened 2020-11-09
Add mpsc_lock_free_queue reals check in reals_ut

dcristoloveanu:
Add mpsc_lock_free_queue reals check in reals_ut


anporumb (approved):
:shipit:

PR #78, opened 2021-09-30
Add support for condition function in set value API

avranju:


mattdurak:
> typedef void(*HASH_TABLE_SKIPPED_SEQ_NO_CB)(void* context, int64_t skipped_sequence_no);

I expect to see `CONDITION_CHECK_CB` somewhere here. And likely should be called `HASH_TABLE_CONDITION_CHECK_CB` to avoid naming collisions

---
Refers to: devdoc/clds_hash_table_requirements.md:30 in a32a324. [](commit_id = a32a324894cf62a9a538be4b1d42d2007616f1d2, deletion_comment = False)

devdoc/clds_sorted_list_requirements.md
```diff
@@ -61,10 +61,18 @@ typedef struct CLDS_SORTED_LIST_TAG* CLDS_SORTED_LIST_HANDLE;
...
+    CLDS_CONDITION_CHECK_ERROR, \
+    CLDS_CONDITION_CHECK_NOT_MET
+
+MU_DEFINE_ENUM(CLDS_CONDITION_CHECK_RESULT, CLDS_CONDITION_CHECK_RESULT_VALUES);
+
 typedef void*(*SORTED_LIST_GET_ITEM_KEY_CB)(void* context, struct CLDS_SORTED_LIST_ITEM_TAG* item);
 typedef int(*SORTED_LIST_KEY_COMPARE_CB)(void* context, void* key1, void* key2);
 typedef void(*SORTED_LIST_ITEM_CLEANUP_CB)(void* context, struct CLDS_SORTED_LIST_ITEM_TAG* item);
 typedef void(*SORTED_LIST_SKIPPED_SEQ_NO_CB)(void* context, int64_t skipped_sequence_no);
+typedef CLDS_CONDITION_CHECK_RESULT (*CONDITION_CHECK_CB)(void* context, void* new_key, void* old_key);
```
mattdurak, line 75: I think this should be `SORTED_LIST_CONDITION_CHECK_CB` and it should return a `SORTED_LIST_CONDITION_CHECK_RESULT`.

devdoc/clds_hash_table_requirements.md
```diff
@@ -100,7 +101,7 @@ MOCKABLE_FUNCTION(, CLDS_HASH_TABLE_INSERT_RESULT, clds_hash_table_insert, CLDS_
...
 MOCKABLE_FUNCTION(, CLDS_HASH_TABLE_DELETE_RESULT, clds_hash_table_delete_key_value, CLDS_HASH_TABLE_HANDLE, clds_hash_table, CLDS_HAZARD_POINTERS_THREAD_HANDLE, clds_hazard_pointers_thread, void*, key, CLDS_HASH_TABLE_ITEM*, value, int64_t*, sequence_number);
 MOCKABLE_FUNCTION(, CLDS_HASH_TABLE_REMOVE_RESULT, clds_hash_table_remove, CLDS_HASH_TABLE_HANDLE, clds_hash_table, CLDS_HAZARD_POINTERS_THREAD_HANDLE, clds_hazard_pointers_thread, void*, key, CLDS_HASH_TABLE_ITEM**, item, int64_t*, sequence_number);
-MOCKABLE_FUNCTION(, CLDS_HASH_TABLE_SET_VALUE_RESULT, clds_hash_table_set_value, CLDS_HASH_TABLE_HANDLE, clds_hash_table, CLDS_HAZARD_POINTERS_THREAD_HANDLE, clds_hazard_pointers_thread, void*, key, CLDS_HASH_TABLE_ITEM*, new_item, CLDS_HASH_TABLE_ITEM**, old_item, int64_t*, sequence_number);
+MOCKABLE_FUNCTION(, CLDS_HASH_TABLE_SET_VALUE_RESULT, clds_hash_table_set_value, CLDS_HASH_TABLE_HANDLE, clds_hash_table, CLDS_HAZARD_POINTERS_THREAD_HANDLE, clds_hazard_pointers_thread, void*, key, CLDS_HASH_TABLE_ITEM*, new_item, CONDITION_CHECK_CB, condition_check_func, void*, condition_check_context, CLDS_HASH_TABLE_ITEM**, old_item, int64_t*, sequence_number);
```
mattdurak, line 104: >clds_hash_table_set_value [](http://example.com/codeflow?start=54&length=25)

Just wondering whether we should have a new `clds_hash_table_conditionally_set_value` function instead of changing the existing function signature with something that will normally be passed as `NULL`...

src/clds_hash_table.c
```diff
@@ -967,9 +973,15 @@ CLDS_HASH_TABLE_SET_VALUE_RESULT clds_hash_table_set_value(CLDS_HASH_TABLE_HANDL
...
+                if (sorted_list_set_value == CLDS_SORTED_LIST_SET_VALUE_CONDITION_NOT_MET)
+                {
+                    /* Codes_SRS_CLDS_HASH_TABLE_04_002: [ If clds_sorted_list_set_value returns CLDS_SORTED_LIST_SET_VALUE_CONDITION_NOT_MET, clds_hash_table_set_value shall fail and return CLDS_HASH_TABLE_SET_VALUE_CONDITION_NOT_MET. ]*/
+                    LogError("Condition not met during set value");
+                    result = CLDS_HASH_TABLE_SET_VALUE_CONDITION_NOT_MET;
+                }
+                else if (sorted_list_set_value != CLDS_SORTED_LIST_SET_VALUE_OK)
                 {
                     /* Codes_SRS_CLDS_HASH_TABLE_01_100: [ If clds_sorted_list_set_value returns any other value, clds_hash_table_set_value shall fail and return CLDS_HASH_TABLE_SET_VALUE_ERROR. ]*/
                     LogError("Cannot set key in sorted list");
```
mattdurak, line 987: >ed list"); [](http://example.com/codeflow?start=52&length=10)

I guess we should consider printing the enum here while making these changes.

src/clds_sorted_list.c
```diff
@@ -1410,6 +1410,62 @@ CLDS_SORTED_LIST_SET_VALUE_RESULT clds_sorted_list_set_value(CLDS_SORTED_LIST_HA
...
                             // we are in a stable state, compare the current item key to our key
                             void* current_item_key = clds_sorted_list->get_item_key_cb(clds_sorted_list->get_item_key_cb_context, (struct CLDS_SORTED_LIST_ITEM_TAG*)current_item);
+
+                            if (condition_check_func != NULL)
+                            {
+                                /* Codes_SRS_CLDS_SORTED_LIST_04_001: [ If condition_check_func is not NULL it shall be called passing condition_check_context and the new and old keys. ]*/
+                                CLDS_CONDITION_CHECK_RESULT condition_check_result = condition_check_func(condition_check_context, new_item_key, current_item_key);
+                                switch (condition_check_result)
+                                {
+                                    case CLDS_CONDITION_CHECK_ERROR:
```
mattdurak, line 1420: > case CLDS_CONDITION_CHECK_ERROR: [](http://example.com/codeflow?start=35&length=33)

should also have default label

src/clds_sorted_list.c
```diff
@@ -1410,6 +1410,62 @@ CLDS_SORTED_LIST_SET_VALUE_RESULT clds_sorted_list_set_value(CLDS_SORTED_LIST_HA
...
+                                    {
+                                        LogError("Condition check failed");
+                                        if (previous_hp != NULL)
+                                        {
+                                            // let go of previous hazard pointer
+                                            clds_hazard_pointers_release(clds_hazard_pointers_thread, previous_hp);
+                                        }
+
+                                        clds_hazard_pointers_release(clds_hazard_pointers_thread, current_item_hp);
```
mattdurak, line 1429: Looks like common code to put in the "if" around line 1464

src/clds_sorted_list.c
```diff
@@ -1410,6 +1410,62 @@ CLDS_SORTED_LIST_SET_VALUE_RESULT clds_sorted_list_set_value(CLDS_SORTED_LIST_HA
...
+
+                                /* Codes_SRS_CLDS_SORTED_LIST_04_004: [ If condition_check_func returns CLDS_CONDITION_CHECK_OK then clds_sorted_list_set_value continues. ]*/
+                                if (condition_check_result != CLDS_CONDITION_CHECK_OK)
+                                {
+                                    restart_needed = false;
+                                    break;
+                                }
+                            }
+
                             int compare_result = clds_sorted_list->key_compare_cb(clds_sorted_list->key_compare_cb_context, new_item_key, current_item_key);
```
mattdurak, line 1469: >int compare_result = clds_sorted_list->key_compare_cb(clds_sorted_list->key_compare_cb_context, new_item_key, current_item_key); [](http://example.com/codeflow?start=28&length=128)

Should the condition check happen after this check? It seems to me (and I don't know all the details of clds), that we only consider adding the value in the branches starting around lines 1473 and 1655.

It looks like the current condition check might be called excessively when searching for where the key belongs, e.g. the branch in line 1711.

Now, it may be a pre-mature optimization to worry about extra evaluations, but I'm also wondering about the actual logic.

What if the condition check fails but the keys don't match? It doesn't make sense to set result to `CLDS_SORTED_LIST_SET_VALUE_CONDITION_NOT_MET` if we end up not matching the key and keep searching.

And that result seems a bit "sticky". For example, we could reach line 1449 for the "wrong" key, set result to "NOT_MET", then go to the "correct" key, reach line 1456, but the result is still "NOT_MET".

So I think that the condition check needs to happen as close to the end of this set value operation as possible.

src/clds_sorted_list.c
```diff
@@ -1410,6 +1410,62 @@ CLDS_SORTED_LIST_SET_VALUE_RESULT clds_sorted_list_set_value(CLDS_SORTED_LIST_HA
...
+                                        break;
+                                    }
+                                }
+
+                                /* Codes_SRS_CLDS_SORTED_LIST_04_004: [ If condition_check_func returns CLDS_CONDITION_CHECK_OK then clds_sorted_list_set_value continues. ]*/
+                                if (condition_check_result != CLDS_CONDITION_CHECK_OK)
+                                {
+                                    restart_needed = false;
+                                    break;
+                                }
```
mattdurak, line 1466: Actually doesn't this skip over the branch in line 1711? Do the tests fail (otherwise seems like missing test coverage if I'm reading this code correctly)?

tests/clds_hash_table_int/clds_hash_table_int.c
```diff
@@ -348,7 +348,7 @@ static int continuous_set_value_thread(void* arg)
         CLDS_HASH_TABLE_ITEM* old_item;
         CLDS_HASH_TABLE_SET_VALUE_RESULT set_result;
         int64_t seq_no;
-        set_result = clds_hash_table_set_value(thread_data->hash_table, thread_data->clds_hazard_pointers_thread, (void*)(uintptr_t)(i + 1), item, &old_item, &seq_no);
+        set_result = clds_hash_table_set_value(thread_data->hash_table, thread_data->clds_hazard_pointers_thread, (void*)(uintptr_t)(i + 1), item, NULL, NULL, &old_item, &seq_no);
```
mattdurak, line 351: >NULL [](http://example.com/codeflow?start=147&length=4)

Are int tests in a separate task?

tests/clds_sorted_list_ut/clds_sorted_list_ut.c
```diff
@@ -3882,7 +3891,115 @@ TEST_FUNCTION(clds_sorted_list_set_value_when_the_key_is_in_the_list_succeeds)
...
+TEST_FUNCTION(clds_sorted_list_set_value_fails_when_condition_check_cb_returns_error)
+{
+    test_clds_sorted_list_set_value_with_condition_check(CLDS_CONDITION_CHECK_ERROR, CLDS_SORTED_LIST_SET_VALUE_ERROR);
+}
+
+/* Tests_SRS_CLDS_SORTED_LIST_04_001: [ If condition_check_func is not NULL it shall be called passing condition_check_context and the new and old keys. ]*/
+/* Tests_SRS_CLDS_SORTED_LIST_04_003: [ If condition_check_func returns CLDS_CONDITION_CHECK_NOT_MET then clds_sorted_list_set_value shall fail and return CLDS_SORTED_LIST_SET_VALUE_CONDITION_NOT_MET. ]*/
+TEST_FUNCTION(clds_sorted_list_set_value_fails_when_condition_check_cb_returns_condition_not_met)
+{
+    test_clds_sorted_list_set_value_with_condition_check(CLDS_CONDITION_CHECK_NOT_MET, CLDS_SORTED_LIST_SET_VALUE_CONDITION_NOT_MET);
```
mattdurak, line 3963: Seems like we need more tests with the condition check, like key not found

devdoc/clds_hash_table_requirements.md
dcristoloveanu: I was the one stating that it is Ok to change it as we don't have a concern about breaking APIs since it is an internal API surface.

Frankly, we can also do an overload as separate function and not change any of the existing usages. I'll leave it for you to consider Raj, if you do an overload I do not mind, but if you feel like not doing it and leave it as is and update the usages I think that will not be that big of a deal either.

devdoc/clds_sorted_list_requirements.md
```diff
@@ -74,8 +82,7 @@ typedef struct CLDS_SORTED_LIST_ITEM_TAG
     volatile_atomic int32_t ref_count;
     SORTED_LIST_ITEM_CLEANUP_CB item_cleanup_callback;
     void* item_cleanup_callback_context;
-    volatile_atomic struct CLDS_SORTED_LIST_ITEM_TAG* next;
-    int64_t seq_no;
+    struct CLDS_SORTED_LIST_ITEM_TAG* volatile_atomic next;
```
dcristoloveanu, line 85: > volatile_atomic next; [](http://example.com/codeflow?start=37&length=22)

Neat

devdoc/clds_sorted_list_requirements.md
```diff
@@ -74,8 +82,7 @@ typedef struct CLDS_SORTED_LIST_ITEM_TAG
     volatile_atomic int32_t ref_count;
     SORTED_LIST_ITEM_CLEANUP_CB item_cleanup_callback;
     void* item_cleanup_callback_context;
-    volatile_atomic struct CLDS_SORTED_LIST_ITEM_TAG* next;
-    int64_t seq_no;
```
dcristoloveanu, line 78: >  int64_t seq_no; [](http://example.com/codeflow?start=2&length=17)

Was this not used. Cool. We'll use a bit less memory

src/clds_hash_table.c
```diff
@@ -920,7 +926,7 @@ CLDS_HASH_TABLE_SET_VALUE_RESULT clds_hash_table_set_value(CLDS_HASH_TABLE_HANDL
 
             do
             {
-                /* Codes_SRS_CLDS_HASH_TABLE_01_103: [ clds_hash_table_set_value shall obtain the sorted list at the bucked corresponding to the hash of the key. ]*/
+                /* Codes_SRS_CLDS_HASH_TABLE_01_103: [ clds_hash_table_set_value shall obtain the sorted list at the bucket corresponding to the hash of the key. ]*/
```
dcristoloveanu, line 929: >bucket [](http://example.com/codeflow?start=117&length=6)

haha

dcristoloveanu:
> - **SRS_CLDS_HASH_TABLE_01_103: [** `clds_hash_table_set_value` shall obtain the sorted list at the bucked corresponding to the hash of the key. **]**

bucket

---
Refers to: devdoc/clds_hash_table_requirements.md:393 in a32a324. [](commit_id = a32a324894cf62a9a538be4b1d42d2007616f1d2, deletion_comment = False)

tests/clds_hash_table_ut/clds_hash_table_ut.c
```diff
@@ -2590,7 +2600,7 @@ TEST_FUNCTION(clds_hash_table_set_value_with_non_NULL_sequence_number_when_a_sta
 /* Tests_SRS_CLDS_HASH_TABLE_01_102: [ If the key is not found in any of the non top level buckets arrays, clds_hash_table_set_value: ]*/
 /* Tests_SRS_CLDS_HASH_TABLE_01_103: [ clds_hash_table_set_value shall obtain the sorted list at the bucked corresponding to the hash of the key. ]*/
```
dcristoloveanu, line 2601: >bucked [](http://example.com/codeflow?start=101&length=6)

bucket

tests/clds_hash_table_ut/clds_hash_table_ut.c
```diff
@@ -2623,7 +2633,13 @@ TEST_FUNCTION(clds_hash_table_set_value_with_empty_hash_table_sets_the_value_on_
     clds_hazard_pointers_destroy(hazard_pointers);
 }
 
-static void test_when_clds_sorted_list_set_value_fails_clds_hash_table_set_value_with_empty_hash_table_fails(CLDS_SORTED_LIST_SET_VALUE_RESULT sorted_list_result)
+/* Tests_SRS_CLDS_HASH_TABLE_01_085: [ clds_hash_table_set_value shall go through all non top level bucket arrays and: ]*/
+/* Tests_SRS_CLDS_HASH_TABLE_01_102: [ If the key is not found in any of the non top level buckets arrays, clds_hash_table_set_value: ]*/
+/* Tests_SRS_CLDS_HASH_TABLE_01_103: [ clds_hash_table_set_value shall obtain the sorted list at the bucked corresponding to the hash of the key. ]*/
```
dcristoloveanu, line 2638: >bucked [](http://example.com/codeflow?start=101&length=6)

bucket

src/clds_sorted_list.c
```diff
@@ -1410,6 +1410,62 @@ CLDS_SORTED_LIST_SET_VALUE_RESULT clds_sorted_list_set_value(CLDS_SORTED_LIST_HA
...
+
+                            if (condition_check_func != NULL)
+                            {
+                                /* Codes_SRS_CLDS_SORTED_LIST_04_001: [ If condition_check_func is not NULL it shall be called passing condition_check_context and the new and old keys. ]*/
+                                CLDS_CONDITION_CHECK_RESULT condition_check_result = condition_check_func(condition_check_context, new_item_key, current_item_key);
+                                switch (condition_check_result)
+                                {
+                                    case CLDS_CONDITION_CHECK_ERROR:
+                                    {
+                                        LogError("Condition check failed");
```
dcristoloveanu, line 1422: >iled"); [](http://example.com/codeflow?start=68&length=7)

Consider printing the enum also

dcristoloveanu: Here and below

src/clds_sorted_list.c
```diff
@@ -1410,6 +1410,62 @@ CLDS_SORTED_LIST_SET_VALUE_RESULT clds_sorted_list_set_value(CLDS_SORTED_LIST_HA
...
+
+                                        /* Codes_SRS_CLDS_SORTED_LIST_04_003: [ If condition_check_func returns CLDS_CONDITION_CHECK_NOT_MET then clds_sorted_list_set_value shall fail and return CLDS_SORTED_LIST_SET_VALUE_CONDITION_NOT_MET. ]*/
+                                        result = CLDS_SORTED_LIST_SET_VALUE_CONDITION_NOT_MET;
+                                        
+                                        break;
+                                    }
+
+                                    case CLDS_CONDITION_CHECK_OK:
+                                    {
+                                        /* Nothing to do. Life continues like normal. */
```
dcristoloveanu, line 1456: >g to do. Life continues like normal. */ [](http://example.com/codeflow?start=49&length=39)

:-)

src/clds_sorted_list.c
dcristoloveanu: I think we have to clear what the condition function means.
What we need is a condition function for replacing items. As in when the new item replaces the old one we want to evaluate that the item being replaced has certain properties (a certain etag value).

So I think this should be spelled in the .md and we should have the check done ONLY when we are replacing an item (compare returns 0).

The way it is now it is a bit weird: we're attempting to call the condition function when walking through the list all the time.

dcristoloveanu:
>         case CHAOS_TEST_ACTION_SET_VALUE_ONLY_IF_EXISTS_WHEN_NODE_NOT_THERE:

We need to add another action in here that sets the key/value with a condition function.

---
Refers to: tests/clds_sorted_list_int/clds_sorted_list_int.c:1524 in a32a324. [](commit_id = a32a324894cf62a9a538be4b1d42d2007616f1d2, deletion_comment = False)

dcristoloveanu:
> TEST_FUNCTION(clds_sorted_list_chaos_knight_test)

We should have some int tests added and the chaos test enhanced with having an action of setting the key/value with a condition function.

I am Ok with a separate task so we break down the PRs, but we should add the int tests before we start changing the mai repo to use the new functionality. The reason is that due to this lib being rather ... multi threaded, we found in the past that it is very easy to introduce an issue which would not show up until way later.

This is why we introduced chaos testing in it.

---
Refers to: tests/clds_sorted_list_int/clds_sorted_list_int.c:1564 in a32a324. [](commit_id = a32a324894cf62a9a538be4b1d42d2007616f1d2, deletion_comment = False)

src/clds_sorted_list.c
dcristoloveanu: I think we should fix this before merging, so waiting for it.

dcristoloveanu (changes requested):
:clock1:

devdoc/clds_sorted_list_requirements.md
avranju: Only an MD update. The header already doesn't have these.

devdoc/clds_sorted_list_requirements.md
avranju: Just syncing the MD with the header. :)

devdoc/clds_hash_table_requirements.md
avranju: Ran a search of `clds_hash_table_set_value` in block store:

```
C:\b>rg --type c --files-with-matches clds_hash_table_set_value
deps\clds\tests\reals\real_clds_hash_table_renames.h
deps\clds\tests\reals\real_clds_hash_table.h
deps\clds\tests\clds_hash_table_ut\clds_hash_table_ut.c
deps\clds\tests\clds_hash_table_int\clds_hash_table_int.c
deps\clds\inc\clds\clds_hash_table.h
deps\clds\src\clds_hash_table.c
block_storage_2\tests\bsi_offload_ut\bsi_offload_ut.c
block_storage_2\tests\bsi_local_ut\bsi_local_ut.c
block_storage_2\src\bsi_local.c
block_storage_2\src\bsi_offload.c
```

There are only 2 source files and 2 test files using `clds_hash_table_set_value` in the block store repo which should be easy enough to fix. From this I think it'd be easier to change the function signature instead of making another new function.

devdoc/clds_sorted_list_requirements.md
avranju: The reason I made this generic sounding is because this is the same type that's used in `clds_hash_table_set_value` as well. I thought it'd be strange for the hash table API to have type names that indicate its a sorted list. Another option is to produce new types each for the sorted list and hash table. Should we do that?

src/clds_hash_table.c
avranju: :)

src/clds_hash_table.c
avranju: You mean the `LogError` should print the enum variant name returned by `clds_sorted_list_set_value`?

src/clds_sorted_list.c
avranju: Is a default needed even we have exhaustively matched all variants? Is the idea to guard against future variants being added that go unhandled?

src/clds_hash_table.c
avranju: Saw Dan's comment later. Will print the enums.

tests/clds_hash_table_ut/clds_hash_table_ut.c
avranju: Fixed in req doc as well.

tests/clds_sorted_list_ut/clds_sorted_list_ut.c
avranju: So, condition check can only return one of:

- `CLDS_CONDITION_CHECK_OK`
- `CLDS_CONDITION_CHECK_ERROR`
- `CLDS_CONDITION_CHECK_NOT_MET`

Wouldn't the key not found case be in block store where our particular implementation of the condition check does the key comparison in addition to the etag matching?

tests/clds_hash_table_int/clds_hash_table_int.c
avranju: I honestly didn't think of int tests. Good point. Will add.

avranju:
Ok. Makes sense. Let me add the int and chaos tests merging this PR.

src/clds_sorted_list.c
avranju: From what I can tell the branch in 1711 is operating on the result of calling `key_compare_cb` which we haven't called yet. I suppose this ties in with your other comment about whether the condition check should happen before/after the key compare call. Let's continue discussion there.

avranju:
> typedef void(*HASH_TABLE_SKIPPED_SEQ_NO_CB)(void* context, int64_t skipped_sequence_no);

I am using the exact same types defined in `clds_sorted_list.h` in `clds_hash_table.h`. Like I said in the other comment we *could* mirror the types in `clds_hash_table.h` as well but I am not seeing the value right now.

---
In reply to: [931539785](https://github.com/Azure/clds/pull/78#issuecomment-931539785) [](http://example.com/codeflow?ancestors=931539785,931539785)

---
Refers to: devdoc/clds_hash_table_requirements.md:30 in a32a324. [](commit_id = a32a324894cf62a9a538be4b1d42d2007616f1d2, deletion_comment = False)

avranju:
> > typedef void(_HASH_TABLE_SKIPPED_SEQ_NO_CB)(void_ context, int64_t skipped_sequence_no);
> 
> I expect to see `CONDITION_CHECK_CB` somewhere here. And likely should be called `HASH_TABLE_CONDITION_CHECK_CB` to avoid naming collisions

I am using the exact same types defined in `clds_sorted_list.h` in `clds_hash_table.h`. Like I said in the other comment we *could* mirror the types in `clds_hash_table.h` as well but I am not seeing the value right now.

src/clds_sorted_list.c
avranju: Makes sense. I suppose some tests are missing since this issue didn't surface. Will fix.

avranju: Moved the condition check to be after the key comparison call.

avranju: I wonder if the condition check function should also receive pointers to the hash table items (old and new) in question?

src/clds_sorted_list.c
dcristoloveanu: yes, I guess there are 2 ways by which we could get to do nothing if we have no default:
- SW bug due to adding a value in the enum and forgetting to handle it.
- The atom from outer space (aka memory corruption) changes the value.

The first case is the one we can address by adding a default and frankly it does not cost us a lot to do it. And we regularly do it in our SW.

dcristoloveanu:
>         case CHAOS_TEST_ACTION_SET_VALUE_SAME_ITEM:

I think you should add some int tests at the level of the hash table also.

It won't hurt!

---
Refers to: tests/clds_hash_table_int/clds_hash_table_int.c:1675 in 802e0e3. [](commit_id = 802e0e3b77c0b5872a5e16bec570ab8a80fa3474, deletion_comment = False)

dcristoloveanu:
>         case CHAOS_TEST_ACTION_SET_VALUE_SAME_ITEM:

And the same chaos int test update of course

---
In reply to: [932852195](https://github.com/Azure/clds/pull/78#issuecomment-932852195) [](http://example.com/codeflow?ancestors=932852195)

---
Refers to: tests/clds_hash_table_int/clds_hash_table_int.c:1675 in 802e0e3. [](commit_id = 802e0e3b77c0b5872a5e16bec570ab8a80fa3474, deletion_comment = False)

src/clds_sorted_list.c
```diff
@@ -1410,10 +1411,65 @@ CLDS_SORTED_LIST_SET_VALUE_RESULT clds_sorted_list_set_value(CLDS_SORTED_LIST_HA
...
+                                            /* Nothing to do. Life continues like normal. */
+                                            break;
+                                        }
+                                    }
+
+                                    /* Codes_SRS_CLDS_SORTED_LIST_04_004: [ If condition_check_func returns CLDS_CONDITION_CHECK_OK then clds_sorted_list_set_value continues. ]*/
+                                    if (condition_check_result != CLDS_CONDITION_CHECK_OK)
+                                    {
+                                        restart_needed = false;
+                                        break;
```
dcristoloveanu, line 1469: I think this is leaking. That is probably why the int tests have failed.

dcristoloveanu: I think that in this case you should treat it as a no match and execute the instructions in line 1712.

Tests ftw!

dcristoloveanu (approved):
:shipit:

src/clds_sorted_list.c
avranju: Ok. In this case the `default` handler would be the same as the handler for `CLDS_CONDITION_CHECK_ERROR` then. Will make the change.

src/clds_sorted_list.c
avranju: Not sure I get this one. Control will go into the `if` condition here only if the condition check call returned something other than `CLDS_CONDITION_CHECK_OK`. What is getting leaked? Yet to look at the int test failure but I think that might be a leak in the chaos test.

PR #160, opened 2023-04-12
Remove SMHasherSupport from build, use c_util instead.

darobs:
Remove adding smhasher to build, replace SMHasherSupport with c_util in target_link_libraries.


dcristoloveanu (approved):
:shipit:

PR #172, opened 2023-07-24
Add linux compiler checks

jebrando:


tests/clds_singly_linked_list_ut/clds_singly_linked_list_ut.c
```diff
@@ -1571,10 +1571,9 @@ TEST_FUNCTION(clds_singly_linked_list_find_result_has_the_ref_count_incremented)
     CLDS_HAZARD_POINTERS_THREAD_HANDLE hazard_pointers_thread = clds_hazard_pointers_register_thread(hazard_pointers);
     CLDS_SINGLY_LINKED_LIST_HANDLE list = clds_singly_linked_list_create(hazard_pointers);
     CLDS_SINGLY_LINKED_LIST_ITEM* item = CLDS_SINGLY_LINKED_LIST_NODE_CREATE(TEST_ITEM, test_item_cleanup_func, (void*)0x4242);
-    CLDS_SINGLY_LINKED_LIST_ITEM* result;
     (void)clds_hazard_pointers_set_reclaim_threshold(hazard_pointers, 1);
     (void)clds_singly_linked_list_insert(list, hazard_pointers_thread, item);
-    result = clds_singly_linked_list_find(list, hazard_pointers_thread, test_item_compare, item);
+    (void)clds_singly_linked_list_find(list, hazard_pointers_thread, test_item_compare, item);
```
darobs, line 1576: If the result isn't checked, what use is the call here?

jebrando: That is a good question, it is getting the item to be used later, so we need to make the call.

jebrando: I think the best thing to do is to check the value.  I will add that.

darobs (approved):
:shipit:

PR #177, opened 2023-09-11
LRU Cache using clds_hash_table, SRW Lock and DList

cyrus-jackson:
Implementation for LRU Cache. The design doc is [here](https://github.com/Azure/clds/blob/master/devdoc/lru_cache_design.md)

`put` :
image

`Eviction logic`: 
image

`get`:
image

The operations to clds_hash_table are done under a lock to make sure there are no race conditions

### DIFF
--- a/tests/reals/real_inactive_hp_thread_queue.c
+++ b/tests/reals/real_inactive_hp_thread_queue.c
@@ -3,9 +3,9 @@
 
 #include "real_gballoc_hl.h"
 #include "real_gballoc_hl_renames.h"
-#include "c_pal/gballoc_hl_redirect.h"
 
 #include "real_inactive_hp_thread_queue_renames.h"
+#include "real_interlocked_renames.h"
 
 #include "c_pal/thandle_ll.h"
 #include "c_util/tqueue_ll.h"


devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,201 @@
...
+
+MU_DEFINE_ENUM(LRU_CACHE_PUT_RESULT, LRU_CACHE_PUT_RESULT_VALUES);
+
+#define LRU_CACHE_EVICT_RESULT_VALUES \
+    LRU_CACHE_EVICT_OK, \
+    LRU_CACHE_EVICT_ERROR
+
+MU_DEFINE_ENUM(LRU_CACHE_EVICT_RESULT, LRU_CACHE_EVICT_RESULT_VALUES);
+
+typedef void(*LRU_CACHE_EVICT_CALLBACK_FUNC)(void* context, LRU_CACHE_EVICT_RESULT cache_evict_status);
```
avranju, line 25: Should we pass a reference to the item being evicted in the callback?

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,201 @@
...
+
+**SRS_LRU_CACHE_13_006: [** `start_sequence_number` shall be allowed to be NULL, in which case no sequence number computations shall be performed. **]**
+
+**SRS_LRU_CACHE_13_007: [** `skipped_seq_no_cb` shall be allowed to be NULL. **]**
+
+**SRS_LRU_CACHE_13_008: [** `skipped_seq_no_cb_context` shall be allowed to be NULL. **]**
+
+**SRS_LRU_CACHE_13_009: [** If `start_sequence_number` is NULL, then `skipped_seq_no_cb` must also be NULL, otherwise `lru_cache_create` shall fail and return NULL. **]**
+
+**SRS_LRU_CACHE_13_010: [** If `capacity` is less than or equals to `0`, then `lru_cache_create` shall fail and return NULL. **]**
```
avranju, line 61: Should we make `capacity` an unsigned value?

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,201 @@
...
+
+### lru_cache_destroy
+
+```c
+void lru_cache_destroy(LRU_CACHE_HANDLE lru_cache);
+```
+
+Frees up `LRU_CACHE_HANDLE`.
+
+**SRS_LRU_CACHE_13_021: [** If `lru_cache` is `NULL`, then `lru_cache_destroy` shall fail. **]**
```
avranju, line 92: Being a function that returns `void` what does "shall fail" mean? Is the idea that we will terminate the process?

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,201 @@
...
+**SRS_LRU_CACHE_13_022: [** `lru_cache_destroy` shall free all resources associated with the `LRU_CACHE_HANDLE`. **]**
+
+
+### lru_cache_put
+
+```c
+MOCKABLE_FUNCTION(, LRU_CACHE_PUT_RESULT, lru_cache_put, LRU_CACHE_HANDLE, lru_handle, void*, key, CLDS_HASH_TABLE_ITEM*, value, int64_t, size, int64_t, seq_no, LRU_CACHE_EVICT_CALLBACK_FUNC, evict_callback, void*, evict_context);
+```
+
+The `lru_cache_put` function is utilized for inserting or updating an item in the Least Recently Used (LRU) cache. If the item already exists in the cache, it is removed and reinserted to maintain the LRU order. Additionally, when the cache is at full capacity, this function triggers eviction by removing the least recently used item to make space for the new item. All the latest items are inserted at the tail of the `doubly_linked_list`. During eviction, the node next to the head (i.e., the least recently used item) is selected and removed from the `clds_hash_table`.
```
avranju, line 103: > If the item already exists in the cache, it is removed and reinserted to maintain the LRU order

I think Achintya had the opinion that we should be able to just update the item in place in the hash table without needing to remove and re-insert it.

avranju: > when the cache is at full capacity

What does `capacity` mean here? I don't suppose we have insight into actual memory use do we? As in, the "items" stored in the hash table are just pointers to us aren't they? Do we know how big the memory is that they are pointing to? Or is `capacity` just a count of items in the hash table? Would such a count based control measure enable us to bound memory use in the cache?

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,201 @@
...
+  - Head -> Node (key2) -> Node (key3) (`key1` is evicted)
+
+
+**SRS_LRU_CACHE_13_023: [** If `lru_handle` is `NULL`, then `lru_cache_put` shall fail and return `LRU_CACHE_PUT_ERROR`. **]**
+
+**SRS_LRU_CACHE_13_024: [** If `key` is `NULL`, then `lru_cache_put` shall fail and return `LRU_CACHE_PUT_ERROR`. **]**
+
+**SRS_LRU_CACHE_13_025: [** If `value` is `NULL`, then `lru_cache_put` shall fail and return `LRU_CACHE_PUT_ERROR`. **]**
+
+**SRS_LRU_CACHE_13_026: [** If `size` is `0`, then `lru_cache_put` shall fail and return `LRU_CACHE_PUT_ERROR`. **]**
```
avranju, line 122: Ok, so we do know what the size of the item is. :)

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,201 @@
...
+
+- **SRS_LRU_CACHE_13_043: [** On success, `evict_callback` is called with the status `LRU_CACHE_EVICT_OK`. **]**
+
+**SRS_LRU_CACHE_13_044: [** `lru_cache_put` shall create LRU Node item to be inserted in the hash table. **]**
+
+**SRS_LRU_CACHE_13_045: [** `lru_cache_put` shall insert the LRU Node item in the hash table by calling `clds_hash_table_insert`. **]**
+
+**SRS_LRU_CACHE_13_046: [** `lru_cache_put` shall acquire the lock in exclusive mode. **]**
+
+**SRS_LRU_CACHE_13_047: [** `lru_cache_put` shall append the node to the tail and update the `current_size` with item `size`. **]**
```
avranju, line 164: Should this be 2 requirements? Append to linked list and increment current size?

tests/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ if(${run_int_tests})
 if(WIN32)
         # this test has a problem on Linux, suspicion of badly written test
         build_test_folder(clds_hash_table_int)
+	    build_test_folder(lru_cache_int)
```
avranju, line 28: Why only Win32 for int test?

src/lru_cache.c
```diff
@@ -0,0 +1,379 @@
+// Licensed under the MIT license.See LICENSE file in the project root for full license information.
```
avranju, line 1: Copyright notice?

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,216 @@
...
+
+#define LRU_CACHE_EVICT_RESULT_VALUES \
+    LRU_CACHE_EVICT_OK, \
+    LRU_CACHE_EVICT_ERROR
+
+MU_DEFINE_ENUM(LRU_CACHE_EVICT_RESULT, LRU_CACHE_EVICT_RESULT_VALUES);
+
+typedef void(*LRU_CACHE_EVICT_CALLBACK_FUNC)(void* context, LRU_CACHE_EVICT_RESULT cache_evict_status);
+
+MOCKABLE_FUNCTION(, LRU_CACHE_HANDLE, lru_cache_create, COMPUTE_HASH_FUNC, compute_hash, KEY_COMPARE_FUNC, key_compare_func, size_t, initial_bucket_size, CLDS_HAZARD_POINTERS_HANDLE, clds_hazard_pointers, volatile_atomic int64_t*, start_sequence_number, HASH_TABLE_SKIPPED_SEQ_NO_CB, skipped_seq_no_cb, void*, skipped_seq_no_cb_context, int64_t, capacity);
```
mattdurak, line 27: >capacity [](http://example.com/codeflow?start=346&length=8)

should this be `capacity_bytes`?

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,216 @@
...
+
+MOCKABLE_FUNCTION(, void, lru_cache_destroy, LRU_CACHE_HANDLE, lru_cache);
+
+MOCKABLE_FUNCTION(, LRU_CACHE_PUT_RESULT, lru_cache_put, LRU_CACHE_HANDLE, lru_handle, void*, key, CLDS_HASH_TABLE_ITEM*, value, int64_t, size, int64_t, seq_no, LRU_CACHE_EVICT_CALLBACK_FUNC, evict_callback, void*, evict_context);
+MOCKABLE_FUNCTION(, CLDS_HASH_TABLE_ITEM*, lru_cache_get, LRU_CACHE_HANDLE, lru_cache, void*, key);
```
mattdurak, line 32: Probably want a way to remove items rather than only waiting for them to get evicted.

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,216 @@
+# `lru_cache` requirements
+
+## Overview
+
+A cache module that deletes the least-recently-used (lru) items. The design for this is [here]()
```
mattdurak, line 5: >[here]() [](http://example.com/codeflow?start=88&length=8)

nit: missing link

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,216 @@
...
+
+### clds_hash_table_create
+
+```c
+MOCKABLE_FUNCTION(, LRU_CACHE_HANDLE, lru_cache_create, COMPUTE_HASH_FUNC, compute_hash, KEY_COMPARE_FUNC, key_compare_func, size_t, initial_bucket_size, CLDS_HAZARD_POINTERS_HANDLE, clds_hazard_pointers, volatile_atomic int64_t*, start_sequence_number, HASH_TABLE_SKIPPED_SEQ_NO_CB, skipped_seq_no_cb, void*, skipped_seq_no_cb_context, int64_t, capacity);
+```
+
+Creates `LRU_CACHE_HANDLE` which holds `clds_hash_table`, `doublylinkedlist` and `SRW_LOCK_HANDLE`.
+
+**SRS_LRU_CACHE_13_001: [** If `compute_hash` is NULL, `lru_cache_create` shall fail and return NULL. **]**
```
mattdurak, line 43: >NULL [](http://example.com/codeflow?start=49&length=4)

missing backticks, here and everywhere

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,216 @@
...
+
+**SRS_LRU_CACHE_13_014: [** `lru_cache_create` shall allocate `CLDS_HAZARD_POINTERS_THREAD_HELPER_HANDLE` by calling `clds_hazard_pointers_thread_helper_create`. **]**
+
+**SRS_LRU_CACHE_13_015: [** `lru_cache_create` shall allocate `clds_hash_table` by calling `clds_hash_table_create`. **]**
+
+**SRS_LRU_CACHE_13_016: [** `lru_cache_create` shall allocate `SRW_LOCK_HANDLE` by calling `srw_lock_create`. **]**
+
+**SRS_LRU_CACHE_13_017: [** `lru_cache_create` shall initialize `head` by calling `DList_InitializeListHead`. **]**
+
+**SRS_LRU_CACHE_13_018: [** `lru_cache_create` shall assign default value of `0` to `current_size` and `capacity`. **]**
```
mattdurak, line 77: >capacity [](http://example.com/codeflow?start=104&length=8)

why would it assign 0 to capacity?

devdoc/lru_cache_requirements.md
mattdurak: > We dont terminate the process (should we?). 

We shouldn't.

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,216 @@
...
+
+```c
+MOCKABLE_FUNCTION(, LRU_CACHE_PUT_RESULT, lru_cache_put, LRU_CACHE_HANDLE, lru_handle, void*, key, CLDS_HASH_TABLE_ITEM*, value, int64_t, size, int64_t, seq_no, LRU_CACHE_EVICT_CALLBACK_FUNC, evict_callback, void*, evict_context);
+```
+
+The `lru_cache_put` function is utilized for inserting or updating an item in the Least Recently Used (LRU) cache. If the item already exists in the cache, the `current_size` is updated first, and then the value is reinserted into the cache to maintain the LRU order and triggers eviction if needed. In case the item is not found, it adds the item to the cache using `add_to_cache_internal` and performs eviction if necessary with `evict_internal`. The eviction process involves updating the cache's current size, removing the least recently used item, and invoking an eviction callback. All the latest items are inserted at the tail of the `doubly_linked_list`. During eviction, the node next to the head (i.e., the least recently used item) is selected and removed from the `clds_hash_table`. It's important to note that the `current_size` may temporarily increase during this process, but eviction ensures the `current_size` is normalized.
```
mattdurak, line 103: >add_to_cache_internal [](http://example.com/codeflow?start=368&length=21)

name here is implementation detail, not needed, can just say "...it adds the item to the cache and performs..."

mattdurak, line 103: >evict_internal [](http://example.com/codeflow?start=432&length=14)

Do you want to spec this out separately (because you talked about maybe making it happen on a worker thread later)? If not, we can remove the name here as well.

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,216 @@
...
+
+LRU Cache (Capacity: 2)
+- put(key1, value1)
+  - Head -> Node (key1)
+- put(key2, value2)
+  - Head -> Node (key1) -> Node (key2)
+- put(key3, value3)
+  - Head -> Node (key2) -> Node (key3) (`key1` is evicted)
+
+Note: The `size` of the value needs to be precalculated in terms of the `capacity` mentioned at the creation of the cache.
```
mattdurak, line 115: Ah I see, size/capacity don't have units, they just need to match (either just a count, or actual size in bytes or something else).

mattdurak, line 115: All of the description here is a bit long. Typically these sorts of explanations belong at top of the document somewhere after the overview in something like a "design details" section.

That section(s) can define the logic of when the size is updated/compared to capacity, how eviction works, what capacity means, etc.

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,216 @@
...
+
+**SRS_LRU_CACHE_13_029: [** `lru_cache_put` shall check hash table for any existence of the value by calling `clds_hash_table_find` on the `key`. **]**
+
+**SRS_LRU_CACHE_13_030: [**  If the `key` is found: **]**
+
+- **SRS_LRU_CACHE_13_070: [** `lru_cache_put` shall update the `current_size` with the new `size` and removes the old value size. **]**
+
+- **SRS_LRU_CACHE_13_064: [** `lru_cache_put` shall create LRU Node item to be updated in the hash table. **]**
+
+- **SRS_LRU_CACHE_13_065: [** `lru_cache_put` shall update the LRU Node item in the hash table by calling `clds_hash_table_set_value`. **]**
```
mattdurak, line 137: This is inherently racey. The item may be removed between the check in 13_029 and the update in 13_065 and we need to handle this.

Also, I thought we said we would want to  to the order of updates as:
1. Update CLDS
2. Update list
3. Update size

And the eviction checks in the reverse order:
1. Check size
2. Get node from list
3. Remove from CLDS

I think that ordering makes more sense because then the eviction will only look at the structures if it knows the size has changed. In the reverse case, eviction is lookin at the LRU list before the new item has actually be inserted.

mattdurak: So maybe we need to instead do this:

1. clds_hash_table_set_value
2. update list
3. Check if we replaced an item
4. Update size (adding new size and removing old if it was replaced).

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,216 @@
...
+
+- **SRS_LRU_CACHE_13_046: [** `lru_cache_put` shall acquire the lock in exclusive mode. **]**
+
+- **SRS_LRU_CACHE_13_047: [** `lru_cache_put` shall append the node to the tail. **]**
+
+- **SRS_LRU_CACHE_13_048: [** `lru_cache_put` shall release the lock in exclusive mode. **]**
+
+- **SRS_LRU_CACHE_13_062: [** `lru_cache_put` shall add the item `size` to the `current_size`. **]**
+
+**SRS_LRU_CACHE_13_037: [** While the `capacity` of the cache is full: **]**
```
mattdurak, line 163: >e `capacity` of the cache is full: * [](http://example.com/codeflow?start=36&length=36)

poetry: the capacity doesn't change so to say "while the capacity is full" doesn't make much sense. Likely want "While the current size of the cache exceeds `capacity`:"

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,216 @@
...
+
+- **SRS_LRU_CACHE_13_040: [** `lru_cache_put` shall acquire the lock in exclusive. **]**
+
+- **SRS_LRU_CACHE_13_038: [** `lru_cache_put` shall get the least used node. **]**
```
mattdurak, line 167: >get the least used node. **] [](http://example.com/codeflow?start=52&length=28)

Maybe this should say something about the head of the LRU list

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,216 @@
...
+
+- **SRS_LRU_CACHE_13_072: [** `lru_cache_put` shall decrement the least used node size from `current_size`. **]**
+
+- **SRS_LRU_CACHE_13_039: [** The least used node is removed from `clds_hash_table` by calling `clds_hash_table_remove`. **]**
+
+- **SRS_LRU_CACHE_13_073: [** `lru_cache_put` shall acquire the lock in exclusive. **]**
+
+- **SRS_LRU_CACHE_13_041: [** `lru_cache_put` shall remove the least used node from the DList by calling `DList_RemoveEntryList`. **]**
+
+- **SRS_LRU_CACHE_13_074: [** `lru_cache_put` shall release the lock in exclusive mode. **]**
```
mattdurak, line 179: We are leaving the node in the list during the clds remove? This seems to have extra complexity and races (two evictions could pull the same node and both try to remove it).

Instead I think we should be removing it around 13_038 (get the LRU node and remove it all in one critical section).

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,216 @@
...
+
+**SRS_LRU_CACHE_13_053: [** `lru_cache_get` shall get `CLDS_HAZARD_POINTERS_THREAD_HANDLE` by calling `clds_hazard_pointers_thread_helper_get_thread`. **]**
+
+**SRS_LRU_CACHE_13_054: [** `lru_cache_get` shall check hash table for any existence of the value by calling `clds_hash_table_find` on the `key`. **]**
+
+**SRS_LRU_CACHE_13_055: [**  If the `key` is found and the node from the `key` is not recently used: **]**
+
+- **SRS_LRU_CACHE_13_056: [** `lru_cache_get` shall acquire the lock in exclusive mode. **]**
+
+- **SRS_LRU_CACHE_13_057: [** `lru_cache_get` shall remove the old value node from `doubly_linked_list` by calling `DList_RemoveEntryList`. **]**
```
mattdurak, line 208: >old value node from `doubly_linked_list` by calling `DList_RemoveEntryList`. **]** [](http://example.com/codeflow?start=63&length=82)

This needs to handle the node getting removed already, likely the node needs some additional state that says whether it has been removed from the list and that needs to be used under the same lock.

inc/clds/lru_cache.h
```diff
@@ -0,0 +1,49 @@
...
+
+#define LRU_CACHE_EVICT_RESULT_VALUES \
+    LRU_CACHE_EVICT_OK, \
+    LRU_CACHE_EVICT_ERROR
+
+MU_DEFINE_ENUM(LRU_CACHE_EVICT_RESULT, LRU_CACHE_EVICT_RESULT_VALUES);
+
+typedef void(*LRU_CACHE_EVICT_CALLBACK_FUNC)(void* context, LRU_CACHE_EVICT_RESULT cache_evict_status, CLDS_HASH_TABLE_ITEM* evicted_value);
+
+MOCKABLE_FUNCTION(, LRU_CACHE_HANDLE, lru_cache_create, COMPUTE_HASH_FUNC, compute_hash, KEY_COMPARE_FUNC, key_compare_func, size_t, initial_bucket_size, CLDS_HAZARD_POINTERS_HANDLE, clds_hazard_pointers, volatile_atomic int64_t*, start_sequence_number, HASH_TABLE_SKIPPED_SEQ_NO_CB, skipped_seq_no_cb, void*, skipped_seq_no_cb_context, uint64_t, capacity);
```
mattdurak, line 37: >volatile_atomic [](http://example.com/codeflow?start=205&length=15)

missing c_pal/interlocked.h include

inc/clds/lru_cache.h
```diff
@@ -0,0 +1,49 @@
...
+#ifndef LRU_CACHE_H
+#define LRU_CACHE_H
+
+#ifdef __cplusplus
+#include <cstdint>
+#else
+#include <stdint.h>
+#endif
+
+#include "umock_c/umock_c_prod.h"
```
mattdurak, line 12: nit: this should be last

inc/clds/lru_cache.h
```diff
@@ -0,0 +1,49 @@
...
+#endif
+
+typedef struct LRU_CACHE_TAG* LRU_CACHE_HANDLE;
+
+#define LRU_CACHE_PUT_RESULT_VALUES \
+    LRU_CACHE_PUT_OK, \
+    LRU_CACHE_PUT_ERROR, \
+    LRU_CACHE_PUT_VALUE_INVALID_SIZE
+
+MU_DEFINE_ENUM(LRU_CACHE_PUT_RESULT, LRU_CACHE_PUT_RESULT_VALUES);
```
mattdurak, line 27: >MU_DEFINE_ENUM [](http://example.com/codeflow?start=0&length=14)

missing macro utils include

tests/lru_cache_int/lru_cache_int.c
```diff
@@ -0,0 +1,513 @@
...
+#include "c_pal/timer.h"
+#include "c_pal/threadapi.h"
+
+#include "clds/clds_hazard_pointers.h"
+
+#include "clds/clds_hash_table.h"
+
+#include "clds/lru_cache.h"
+
+#define XTEST_FUNCTION(A) void A(void)
```
mattdurak, line 33: >XTEST_FUNCTION [](http://example.com/codeflow?start=8&length=14)

should probably be removed

tests/lru_cache_int/lru_cache_int.c
```diff
@@ -0,0 +1,513 @@
...
+    for (int i = 0; i < n; i++)
+    {
+        CLDS_HASH_TABLE_NODE_RELEASE(TEST_ITEM, items[i]);
+    }
+
+    lru_cache_destroy(lru_cache);
+    clds_hazard_pointers_destroy(hazard_pointers);
+}
+
+END_TEST_SUITE(TEST_SUITE_NAME_FROM_CMAKE)
```
mattdurak, line 513: are chaos tests coming in a separate PR? There are a lot of potential races here that we should have tests for.

src/lru_cache.c
```diff
@@ -0,0 +1,455 @@
...
+            LogError("malloc failed for LRU_CACHE_HANDLE");
+            result = NULL;
+        }
+        else
+        {
+            /*Codes_SRS_LRU_CACHE_13_005: [ start_sequence_number shall be used as the sequence number variable that shall be incremented at every operation that is done on the hash table. ]*/
+            /*Codes_SRS_LRU_CACHE_13_012: [ lru_cache_create shall assign start_sequence_number to seq_no if not NULL, otherwise seq_no is defaulted to 0. ]*/
+            if (start_sequence_number == NULL)
+            {
+                interlocked_exchange_64(lru_cache->seq_no, 0);
```
mattdurak, line 92: >interlocked_exchange_64(lru_cache->seq_no, 0); [](http://example.com/codeflow?start=16&length=46)

Looks like an uninitialized pointer. If we really want to make it optional, then this structure needs a `int64_t` field backing this.

But do we need to make it optional? Can we decide if we need the sequence number and require it, or don't need it and don't even take it as a parameter?

src/lru_cache.c
```diff
@@ -0,0 +1,455 @@
...
+                if (lru_cache->table == NULL)
+                {
+                    /*Codes_SRS_LRU_CACHE_13_020: [ If there are any failures then lru_cache_create shall fail and return NULL. ]*/
+                    LogError("Cannot create table, clds_hash_table_create failed");
+                    result = NULL;
+                }
+                else
+                {
+                    /*Codes_SRS_LRU_CACHE_13_016: [ lru_cache_create shall allocate SRW_LOCK_HANDLE by calling srw_lock_create. ]*/
+                    lru_cache->lock = srw_lock_create(false, "lru_cache_lock");
```
mattdurak, line 123: >srw_lock_create [](http://example.com/codeflow?start=38&length=15)

consider srw_lock_ll if we don't care about statistics (since we are passing false).

src/lru_cache.c
```diff
@@ -0,0 +1,455 @@
...
+    {
+        /*Codes_SRS_LRU_CACHE_13_021: [ If lru_cache is NULL, then lru_cache_destroy shall fail. ]*/
+        LogError("Invalid arguments: LRU_CACHE_HANDLE lru_cache=%p", lru_cache);
+    }
+    else
+    {
+        /*Codes_SRS_LRU_CACHE_13_022: [ lru_cache_destroy shall free all resources associated with the LRU_CACHE_HANDLE. ]*/
+        srw_lock_destroy(lru_cache->lock);
+        clds_hash_table_destroy(lru_cache->table);
+        //clds_hazard_pointers_destroy(lru_cache->clds_hazard_pointers);
```
mattdurak, line 166: >//clds_hazard_pointers_destroy(lru_cache->clds_hazard_pointers); [](http://example.com/codeflow?start=8&length=64)

remove

dcristoloveanu:
> 1. The table contains `LRU_NODE` instances as the value corresponding to the supplied keys. The `LRU_NODE` contains a `DLIST_ENTRY` which points to a node in the `doubly_linked_list`. A pointer to the `LRU_NODE` is stored in the `doubly_linked_list` as well.

technically it does not point - as in it does not have a pointer to an item. It is an anchor for doubly linked list purposes, because it is part of a structure that the doubly linked list does not know about.

---
Refers to: devdoc/lru_cache_design.md:41 in 706a111. [](commit_id = 706a111b1018b69db6130315189ab04a5e7cd6a4, deletion_comment = False)

dcristoloveanu:
> 1. The table contains `LRU_NODE` instances as the value corresponding to the supplied keys. The `LRU_NODE` contains a `DLIST_ENTRY` which points to a node in the `doubly_linked_list`. A pointer to the `LRU_NODE` is stored in the `doubly_linked_list` as well.

I think this is creating confusion. It is by the use of the DLIST_ENTRY that the nodes can be tracked in the DLIST.

---
Refers to: devdoc/lru_cache_design.md:41 in 706a111. [](commit_id = 706a111b1018b69db6130315189ab04a5e7cd6a4, deletion_comment = False)

devdoc/lru_cache_design.md
```diff
@@ -43,61 +43,81 @@ LRU Cache uses a `clds_hash_table`, a `srw_lock`, and a `doubly_linked_list`.
...
+    N --> O
+    H --> O
+    O --> Z
+
 ```
 
 ### Eviction Logic
 
-To evict the least recently used item from the cache, the key from the tail of the `doubly_linked_list` is removed from the `table`, and the node is removed from the `doubly_linked_list`. When an item gets evicted, the provided callback is triggered to report the eviction status. This is done in a loop until there is enough space in the cache.
+The eviction process involves updating the cache's current size, removing the least recently used item, and invoking an eviction callback. All the latest items are inserted at the tail of the `doubly_linked_list`. During eviction, the node next to the head (i.e., the least recently used item) is selected and removed from the `clds_hash_table`. This is done in a loop until there is enough space in the cache. It's important to note that the `current_size` may temporarily increase during this process, but eviction ensures the `current_size` is normalized.
```
dcristoloveanu, line 88: >t`. Dur [](http://example.com/codeflow?start=210&length=7)

Can we please lean on the keyboard and get some CRLFs between paragraphs? Makes it easier to read the .md :-).

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,204 @@
...
+
+#define LRU_CACHE_EVICT_RESULT_VALUES \
+    LRU_CACHE_EVICT_OK, \
+    LRU_CACHE_EVICT_ERROR
+MU_DEFINE_ENUM(LRU_CACHE_EVICT_RESULT, LRU_CACHE_EVICT_RESULT_VALUES);
+
+
+typedef void(*LRU_CACHE_EVICT_CALLBACK_FUNC)(void* context, LRU_CACHE_EVICT_RESULT cache_evict_status, void* evicted_value);
+
+MOCKABLE_FUNCTION(, LRU_CACHE_HANDLE, lru_cache_create, COMPUTE_HASH_FUNC, compute_hash, KEY_COMPARE_FUNC, key_compare_func, size_t, initial_bucket_size, CLDS_HAZARD_POINTERS_HANDLE, clds_hazard_pointers, uint64_t, capacity);
```
dcristoloveanu, line 26: >size_t [](http://example.com/codeflow?start=125&length=6)

Can we use uint32_t?

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,204 @@
...
+
+
+### lru_cache_put
+
+```c
+MOCKABLE_FUNCTION(, LRU_CACHE_PUT_RESULT, lru_cache_put, LRU_CACHE_HANDLE, lru_handle, void*, key, void*, value, uint64_t, size, LRU_CACHE_EVICT_CALLBACK_FUNC, evict_callback, void*, evict_context);
+```
+
+The `lru_cache_put` function is utilized for inserting or updating an item in the Least Recently Used (LRU) cache. If the item already exists in the cache, the `current_size` is updated first, and then the value is reinserted into the cache to maintain the LRU order and triggers eviction if needed. In case the item is not found, it adds the item to the cache  and performs eviction if necessary. The eviction process involves updating the cache's current size, removing the least recently used item, and invoking an eviction callback. All the latest items are inserted at the tail of the `doubly_linked_list`. During eviction, the node next to the head (i.e., the least recently used item) is selected and removed from the `clds_hash_table`. It's important to note that the `current_size` may temporarily increase during this process, but eviction ensures the `current_size` is normalized.
```
dcristoloveanu, line 91: >he  and p [](http://example.com/codeflow?start=358&length=9)

Extra space

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,204 @@
...
+
+**SRS_LRU_CACHE_13_011: [** `lru_cache_create` shall allocate memory for `LRU_CACHE_HANDLE`. **]**
+
+**SRS_LRU_CACHE_13_013: [** `lru_cache_create` shall assign `clds_hazard_pointers` to `LRU_CACHE_HANDLE`. **]**
+
+**SRS_LRU_CACHE_13_014: [** `lru_cache_create` shall allocate `CLDS_HAZARD_POINTERS_THREAD_HELPER_HANDLE` by calling `clds_hazard_pointers_thread_helper_create`. **]**
+
+**SRS_LRU_CACHE_13_015: [** `lru_cache_create` shall allocate `clds_hash_table` by calling `clds_hash_table_create`. **]**
+
+**SRS_LRU_CACHE_13_016: [** `lru_cache_create` shall allocate `SRW_LOCK_LL` by calling `srw_lock_ll_init`. **]**
```
dcristoloveanu, line 61: >allocate [](http://example.com/codeflow?start=53&length=8)

The proper term would be initialize since it is a SRW_LOCK_LL

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,204 @@
...
+
+**SRS_LRU_CACHE_13_017: [** `lru_cache_create` shall initialize `head` by calling `DList_InitializeListHead`. **]**
```
dcristoloveanu, line 63: > `head` [](http://example.com/codeflow?start=63&length=7)

Consider saying "initialize the doubly linked list". head is very specific with regards to the field used in the code.

devdoc/lru_cache_requirements.md
```diff
@@ -0,0 +1,204 @@
...
+
+**SRS_LRU_CACHE_13_018: [** `lru_cache_create` shall assign default value of `0` to `current_size` and the capacity to `capacity`. **]**
```
dcristoloveanu, line 65: >default [](http://example.com/codeflow?start=60&length=7)

I think default does not add any value here. assign 0 is very clear.

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include <stdlib.h>
+#include <inttypes.h>
+#include <stdbool.h>
```
dcristoloveanu, line 6: Not needed

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+
+#include "c_logging/logger.h"
+
+#include "c_pal/gballoc_hl.h"
+#include "c_pal/gballoc_hl_redirect.h"
+#include "c_pal/sync.h"
+#include "c_pal/interlocked.h"
+
+#include "c_util/rc_string.h"
```
dcristoloveanu, line 15: Not needed

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+#include "c_pal/thandle.h"
```
dcristoloveanu, line 16: Also not needed

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+#include "c_pal/srw_lock_ll.h"
+
+#include "c_util/doublylinkedlist.h"
+
```
dcristoloveanu, line 20: Nitpick: Usually we go in the order
c_logging
c_pal
c_util
clds
...

These seem to be a bit intertwined.

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+typedef struct LRU_CACHE_TAG
+{
+    CLDS_HAZARD_POINTERS_HANDLE clds_hazard_pointers;
+    CLDS_HAZARD_POINTERS_THREAD_HELPER_HANDLE clds_hazard_pointers_thread_helper;
+
+    CLDS_HASH_TABLE_HANDLE table;
+
+    volatile_atomic int64_t current_size;
+    uint64_t capacity;
```
dcristoloveanu, line 43: >uint64_t [](http://example.com/codeflow?start=4&length=8)

I find it weird that capacity is uint64_t and size is int64_t. Isn't there a gap?

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+    SRW_LOCK_LL srw_lock;
+} LRU_CACHE;
+
+typedef struct LRU_NODE_TAG
+{
+    void* key;
+    int64_t size;
+    CLDS_HASH_TABLE_ITEM* value;
+    DLIST_ENTRY node;
+    volatile_atomic int32_t state; /* LRU_NODE_STATE */
```
dcristoloveanu, line 56: >/* LRU_NODE_STATE */ [](http://example.com/codeflow?start=35&length=20)

Nice

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+                    result = NULL;
+                }
+                else
+                {
+                    /*Codes_SRS_LRU_CACHE_13_016: [ lru_cache_create shall allocate SRW_LOCK_LL by calling srw_lock_ll_init. ]*/
+                    if (srw_lock_ll_init(&lru_cache->srw_lock) != 0)
+                    {
+                        /*Codes_SRS_LRU_CACHE_13_020: [ If there are any failures then lru_cache_create shall fail and return NULL. ]*/
+                        LogError("failure in srw_lock_ll_init(&lru_cache->srw_lock);");
+                        result = NULL;
```
dcristoloveanu, line 120: > result = NULL; [](http://example.com/codeflow?start=23&length=15)

I think you can simply have result = NULL between lines 140 and 141 and then it is not repeated in each if.

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+        {
+            /*Codes_SRS_LRU_CACHE_13_047: [ lru_cache_put shall append the node to the tail ]*/
+            DList_InitializeListHead(&(new_node->node));
+            DList_InsertTailList(&(lru_cache->head), &(new_node->node));
+        }
+        /*Codes_SRS_LRU_CACHE_13_048: [ lru_cache_put shall release the lock in exclusive mode. ]*/
+        srw_lock_ll_release_exclusive(&lru_cache->srw_lock);
+
+        /*Codes_SRS_LRU_CACHE_13_062: [ lru_cache_put shall add the item size to the current_size. ]*/
+        (void)interlocked_add_64(&lru_cache->current_size, size);
```
dcristoloveanu, line 196: >ache->current_size, size); [](http://example.com/codeflow?start=39&length=26)

Can this overflow?

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+}
+
+
+static LRU_CACHE_EVICT_RESULT evict_internal(LRU_CACHE_HANDLE lru_cache, CLDS_HAZARD_POINTERS_THREAD_HANDLE hazard_pointers_thread, LRU_CACHE_EVICT_CALLBACK_FUNC evict_callback, void* context)
+{
+    LRU_CACHE_EVICT_RESULT result = LRU_CACHE_EVICT_OK;
+    do
+    {
+        int64_t current_size = interlocked_add_64(&lru_cache->current_size, 0);
+        if (result != LRU_CACHE_EVICT_OK || (uint64_t)current_size <= lru_cache->capacity)
```
dcristoloveanu, line 208: >(uint64_t) [](http://example.com/codeflow?start=44&length=10)

Meh, this is what I was talking about with the difference in types a bit earlier.

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+                LogError("Something is wrong. The cache is empty but there is no capacity");
+                srw_lock_ll_release_exclusive(&lru_cache->srw_lock);
+                evict_callback(context, LRU_CACHE_EVICT_ERROR, NULL);
+                result = LRU_CACHE_EVICT_ERROR;
+                break;
+            }
+
+            /*Codes_SRS_LRU_CACHE_13_038: [ lru_cache_put shall get the least used node which is Flink of head node. ]*/
+            DLIST_ENTRY* least_used_node = lru_cache->head.Flink;
+            LRU_NODE* least_used_node_value = (LRU_NODE*)CONTAINING_RECORD(least_used_node, LRU_NODE, node);
```
dcristoloveanu, line 229: >(LRU_NODE*) [](http://example.com/codeflow?start=46&length=11)

What happens if you remove the cast?

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+
+            /*Codes_SRS_LRU_CACHE_13_042: [ lru_cache_put shall release the lock in exclusive mode. ]*/
+            srw_lock_ll_release_exclusive(&lru_cache->srw_lock);
+
+            /*Codes_SRS_LRU_CACHE_13_072: [ lru_cache_put shall decrement the least used node size from current_size. ]*/
+            if (interlocked_compare_exchange_64(&lru_cache->current_size, current_size - least_used_node_value->size, current_size) != current_size)
```
dcristoloveanu, line 235: >current_size - least_used_node_value->size [](http://example.com/codeflow?start=74&length=42)

Can this underflow? We should check/have tests for it.

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+                        {
+                            int32_t state = interlocked_add(&least_used_node_value->state, 0);
+                            if (state != LRU_NODE_STATE_READY)
+                            {
+                                LogWarning("invalid state(%" PRI_MU_ENUM ") for lru node. Node has already been evicted.", MU_ENUM_VALUE(LRU_NODE_STATE, state));
+                            }
+                            else
+                            {
+                                /*Codes_SRS_LRU_CACHE_13_041: [ If LRU Node state is LRU_NODE_STATE_READY only then the old Node is removed from list by calling DList_RemoveEntryList. ]*/
+                                (void)DList_RemoveEntryList(least_used_node);
```
dcristoloveanu, line 262: >       (void)DList_RemoveEntryList(least_used_node); [](http://example.com/codeflow?start=25&length=52)

Somewhere around here we should have a LogVerbose

dcristoloveanu: Otherwise figuring out what happened in PROD will be tough.

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+                /*Codes_SRS_LRU_CACHE_13_050: [ For any other errors, lru_cache_put shall return LRU_CACHE_PUT_ERROR ]*/
+                LogError("Something is wrong. The cache is empty but there is no capacity");
+                srw_lock_ll_release_exclusive(&lru_cache->srw_lock);
+                evict_callback(context, LRU_CACHE_EVICT_ERROR, NULL);
+                result = LRU_CACHE_EVICT_ERROR;
+                break;
+            }
+
+            /*Codes_SRS_LRU_CACHE_13_038: [ lru_cache_put shall get the least used node which is Flink of head node. ]*/
+            DLIST_ENTRY* least_used_node = lru_cache->head.Flink;
```
dcristoloveanu, line 228: >least_used_node [](http://example.com/codeflow?start=25&length=15)

This least used node is saved here under the lock.
And then it is later used in ln 262.
I don't think it can be absolutely guaranteed that it has not changed, can it?

dcristoloveanu: This smells like some AV waiting to happen.

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+            LRU_NODE* least_used_node_value = (LRU_NODE*)CONTAINING_RECORD(least_used_node, LRU_NODE, node);
+
+            /*Codes_SRS_LRU_CACHE_13_042: [ lru_cache_put shall release the lock in exclusive mode. ]*/
+            srw_lock_ll_release_exclusive(&lru_cache->srw_lock);
+
```
dcristoloveanu, line 233: Between ln 216 and 232 there is no modification done, we're only reading the node (questionable if that's all we should do). But if we only read, why would we acquire the lock exclusively?

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+            {
+                /*Codes_SRS_LRU_CACHE_13_050: [ For any other errors, lru_cache_put shall return LRU_CACHE_PUT_ERROR ]*/
+                LogError("clds_hazard_pointers_thread_helper_get_thread failed.");
+                result = LRU_CACHE_PUT_ERROR;
+            }
+
+            /*Codes_SRS_LRU_CACHE_13_029: [ lru_cache_put shall check hash table for any existence of the value by calling clds_hash_table_find on the key. ]*/
+            CLDS_HASH_TABLE_ITEM* hash_table_item = clds_hash_table_find(lru_cache->table, hazard_pointers_thread, key);
+            if (hash_table_item != NULL)
+            {
```
dcristoloveanu, line 342: I don't think we should proceed with executing with hazard_pointers_thread being NULL.

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+                        LogWarning("invalid state(%" PRI_MU_ENUM ") for lru node. Node has already been evicted.", MU_ENUM_VALUE(LRU_NODE_STATE, state));
+                    }
+                    else
+                    {
+                        /*Codes_SRS_LRU_CACHE_13_077: [ If LRU Node state is LRU_NODE_STATE_READY only then the old Node is removed from list by calling DList_RemoveEntryList. ]*/
+                        DList_RemoveEntryList(node);
+                    }
+
+                    DList_InitializeListHead(&(new_node->node));
+
```
dcristoloveanu, line 383: Why is this needed? I think it is not. And has no spec. I suspect that you wrote the code first?

If that is the case, we should discuss :-). I'd like to clear the air around why we write tests first.

dcristoloveanu: No shaming or anything, good topic for 1:1 :-D

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+                if (state != LRU_NODE_STATE_READY)
+                {
+                    /*Codes_SRS_LRU_CACHE_13_079: [ If the LRU Node is already evicted with the state LRU_NODE_STATE_EVICTED, then lru_cache_get shall return CLDS_HASH_TABLE_ITEM value of the key. ]*/
+                    LogWarning("invalid state(%" PRI_MU_ENUM "). The node has already been evicted.", MU_ENUM_VALUE(LRU_NODE_STATE, state));
+                }
+                else
+                {
+                    PDLIST_ENTRY node = &(current_item->node);
+                    /*Codes_SRS_LRU_CACHE_13_056: [ lru_cache_get shall acquire the lock in exclusive mode. ]*/
+                    srw_lock_ll_acquire_exclusive(&lru_cache->srw_lock);
```
dcristoloveanu, line 459: >srw_lock_ll_acquire_exclusive [](http://example.com/codeflow?start=20&length=29)

I don't think we should acquire exclusive for a get operation.

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+                LogError("Eviction failed.");
+            }
+        }
+    }
+    /*Codes_SRS_LRU_CACHE_13_049: [ On success, lru_cache_put shall return LRU_CACHE_PUT_OK. ]*/
+    return result;
+}
+
+
+void* lru_cache_get(LRU_CACHE_HANDLE lru_cache, void* key)
```
dcristoloveanu, line 419: >void* [](http://example.com/codeflow?start=0&length=5)

While looking at this I thought: wouldn't this be nice to be templatized :-)? A wild consider obviously.

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+        {
+            /*Codes_SRS_LRU_CACHE_13_061: [ If there are any failures, lru_cache_get shall return NULL. ]*/
+            LogError("clds_hazard_pointers_thread_helper_get_thread failed");
+            result = NULL;
+        }
+        else
+        {
+            /*Codes_SRS_LRU_CACHE_13_054: [ lru_cache_get shall check hash table for any existence of the value by calling clds_hash_table_find on the key. ]*/
+            CLDS_HASH_TABLE_ITEM* hash_table_item = clds_hash_table_find(lru_cache->table, hazard_pointers_thread, key);
+            if (hash_table_item != NULL)
```
dcristoloveanu, line 446: >  if (hash_table_item != NULL) [](http://example.com/codeflow?start=10&length=30)

On the else branch we should probably LogVerbose.

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+                {
+                    /*Codes_SRS_LRU_CACHE_13_079: [ If the LRU Node is already evicted with the state LRU_NODE_STATE_EVICTED, then lru_cache_get shall return CLDS_HASH_TABLE_ITEM value of the key. ]*/
+                    LogWarning("invalid state(%" PRI_MU_ENUM "). The node has already been evicted.", MU_ENUM_VALUE(LRU_NODE_STATE, state));
+                }
+                else
+                {
+                    PDLIST_ENTRY node = &(current_item->node);
+                    /*Codes_SRS_LRU_CACHE_13_056: [ lru_cache_get shall acquire the lock in exclusive mode. ]*/
+                    srw_lock_ll_acquire_exclusive(&lru_cache->srw_lock);
+                    {
```
dcristoloveanu, line 460: > { [](http://example.com/codeflow?start=19&length=2)

Likely not needed

src/lru_cache.c
```diff
@@ -0,0 +1,483 @@
...
+                        /*Codes_SRS_LRU_CACHE_13_055: [ If the key is found and the node from the key is not recently used: ]*/
+                        if (lru_cache->head.Blink != node)
+                        {
+                            /*Codes_SRS_LRU_CACHE_13_057: [ lru_cache_get shall remove the old value node from doubly_linked_list by calling DList_RemoveEntryList. ]*/
+                            DList_RemoveEntryList(node);
+                            DList_InitializeListHead(node);
```
dcristoloveanu, line 466: Highly unnecessary

tests/lru_cache_int/lru_cache_int.c
```diff
@@ -0,0 +1,489 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include <stdlib.h>
+#include <inttypes.h>
+#include <stdbool.h>
```
dcristoloveanu, line 6: Not needed

tests/lru_cache_int/lru_cache_int.c
```diff
@@ -0,0 +1,489 @@
...
+TEST_FUNCTION_CLEANUP(method_cleanup)
+{
+}
+
+typedef struct EVICT_CONTEXT_TAG
+{
+    uint32_t count;
+} EVICT_CONTEXT;
+
+void on_evict_callback(void* context, LRU_CACHE_EVICT_RESULT cache_evict_status, void* item)
```
dcristoloveanu, line 110: >on_evict_callback [](http://example.com/codeflow?start=5&length=17)

Likely static

tests/lru_cache_int/lru_cache_int.c
```diff
@@ -0,0 +1,489 @@
...
+    // cleanup
+    CLDS_HASH_TABLE_NODE_RELEASE(TEST_ITEM, item1);
+    CLDS_HASH_TABLE_NODE_RELEASE(TEST_ITEM, item2);
+
+    lru_cache_destroy(lru_cache);
+    clds_hazard_pointers_destroy(hazard_pointers);
+}
+
+
+TEST_FUNCTION(test_put_same_key_calls_evict_to_make_space)
```
dcristoloveanu, line 433: >test_put_same_key_calls_evict_to_make_space [](http://example.com/codeflow?start=14&length=43)

We have to add a chaos int test.
No way we should ship this new cache without that. Keep a counter of what it finds :-).

tests/lru_cache_int/lru_cache_int.c
```diff
@@ -0,0 +1,489 @@
...
+{
+    // arrange
+    EVICT_CONTEXT count_context = { 0 };
+
+    CLDS_HAZARD_POINTERS_HANDLE hazard_pointers = clds_hazard_pointers_create();
+    ASSERT_IS_NOT_NULL(hazard_pointers);
+    uint64_t capacity = 11;
+    int n = 10;
+    CLDS_HASH_TABLE_ITEM** items = malloc(n * sizeof(CLDS_HASH_TABLE_ITEM*));
```
dcristoloveanu, line 442: >malloc [](http://example.com/codeflow?start=35&length=6)

This should be malloc_2

tests/lru_cache_int/lru_cache_int.c
```diff
@@ -0,0 +1,489 @@
...
+
+    LRU_CACHE_HANDLE lru_cache = lru_cache_create(test_compute_hash, test_key_compare, 1, hazard_pointers, capacity);
+    ASSERT_IS_NOT_NULL(lru_cache);
+
+    LRU_CACHE_PUT_RESULT result;
+
+    for (int i = 0; i < n; i++)
+    {
+        items[i] = CLDS_HASH_TABLE_NODE_CREATE(TEST_ITEM, NULL, NULL);
```
dcristoloveanu, line 451: >items[i] [](http://example.com/codeflow?start=8&length=8)

Missing ASSERT_IS_NOT_NULL

tests/lru_cache_int/lru_cache_int.c
```diff
@@ -0,0 +1,489 @@
...
+        TEST_ITEM* test_item1 = CLDS_HASH_TABLE_GET_VALUE(TEST_ITEM, items[i]);
+        test_item1->key = i + 1;
+        test_item1->appendix = i + 10;
+
+        result = lru_cache_put(lru_cache, (void*)(uintptr_t)(i+1), items[i], 1, on_evict_callback, &count_context);
+        ASSERT_ARE_EQUAL(LRU_CACHE_PUT_RESULT, LRU_CACHE_PUT_OK, result);
+    }
+
+    CLDS_HASH_TABLE_ITEM* large_item = CLDS_HASH_TABLE_NODE_CREATE(TEST_ITEM, NULL, NULL);
```
dcristoloveanu, line 460: >large_item [](http://example.com/codeflow?start=26&length=10)

Missing ASSERT_IS_NOT_NULL.

tests/lru_cache_int/lru_cache_int.c
```diff
@@ -0,0 +1,489 @@
...
+    // arrange
+    EVICT_CONTEXT count_context = { 0 };
+
+    CLDS_HAZARD_POINTERS_HANDLE hazard_pointers = clds_hazard_pointers_create();
+    ASSERT_IS_NOT_NULL(hazard_pointers);
+    LRU_CACHE_HANDLE lru_cache = lru_cache_create(test_compute_hash, test_key_compare, 1, hazard_pointers, 3);
+    ASSERT_IS_NOT_NULL(lru_cache);
+
+    LRU_CACHE_PUT_RESULT result;
+    CLDS_HASH_TABLE_ITEM* item1 = CLDS_HASH_TABLE_NODE_CREATE(TEST_ITEM, NULL, NULL);
```
dcristoloveanu, line 389: >item1 = CLDS_HASH_TABL [](http://example.com/codeflow?start=26&length=22)

Missing NULL assert. Here and everywhere.

tests/lru_cache_ut/lru_cache_ut.c
```diff
@@ -0,0 +1,1522 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include <stdbool.h>
```
dcristoloveanu, line 4: Not needed. This and the umock ones for bool.

tests/lru_cache_ut/lru_cache_ut.c
```diff
@@ -0,0 +1,1522 @@
...
+
+    set_lru_create_expectations(bucket_size, test_clds_hazard_pointers);
+
+    // act
+    lru_cache = lru_cache_create(test_compute_hash, test_key_compare_func, bucket_size, test_clds_hazard_pointers, capacity);
+
+    // assert
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+    ASSERT_IS_NOT_NULL(lru_cache);
+
```
dcristoloveanu, line 310: Usually it is easier to put this before checking the expected/actual calls. My experience at least. /consider.

dcristoloveanu:
I would like to see the next iteration because 2 items (which will require a push of a new commit anyway)
- Clearing the acquire_exclusive, get a pointer, release, acquire again, use the previous saved pointer issue in evict. I think releasing the lock and using a pointer that was saved is dangerous.
- chaos int test

dcristoloveanu (changes requested):
:clock1:

src/lru_cache.c
```diff
@@ -0,0 +1,505 @@
...
+                result = LRU_CACHE_PUT_ERROR;
+            }
+            else
+            {
+                /*Codes_SRS_LRU_CACHE_13_029: [ lru_cache_put shall check hash table for any existence of the value by calling clds_hash_table_find on the key. ]*/
+                CLDS_HASH_TABLE_ITEM* hash_table_item = clds_hash_table_find(lru_cache->table, hazard_pointers_thread, key);
+                if (hash_table_item != NULL)
+                {
+                    /*Codes_SRS_LRU_CACHE_13_033: [ lru_cache_put shall acquire the lock in exclusive mode. ]*/
+                    srw_lock_ll_acquire_exclusive(&lru_cache->srw_lock);
```
mattdurak, line 360: What if the item is evicted between the find and the lock acquire?

Can we avoid the find altogether and just always do `clds_hash_table_set_value` (and check if we have an `old_value` or not)?

mattdurak: You have the same issue with the else branch where we might not find the item, then it gets inserted on another thread before we take the lock here.

src/lru_cache.c
```diff
@@ -0,0 +1,505 @@
...
+            }
+        }
+        /*Codes_SRS_LRU_CACHE_13_042: [ lru_cache_put shall release the lock in exclusive mode. ]*/
+        srw_lock_ll_release_exclusive(&lru_cache->srw_lock);
+    } while (1);
+
+    return result;
+}
+
+LRU_CACHE_PUT_RESULT lru_cache_put(LRU_CACHE_HANDLE lru_cache, void* key, void* value, int64_t size, LRU_CACHE_EVICT_CALLBACK_FUNC evict_callback, void* context)
```
mattdurak, line 315: >LRU_CACHE_EVICT_CALLBACK_FUNC evict_callback, void* context [](http://example.com/codeflow?start=101&length=59)

I'm reading this code more and now I'm confused. What is `evict_callback` meant to do?

My expectation: I pass `evict_callback` when I put an item with a key/value (call it X). When *that* item X is evicted some time later.

Apparent reality on line 430: I pass `evict_callback` when I put an item with a key/value (X) and if *that* put call causes some *other* item Y to be evicted, then my callback is called.

Why would the caller of put care about other items getting evicted?

src/lru_cache.c
```diff
@@ -0,0 +1,505 @@
...
+    SRW_LOCK_LL srw_lock;
+} LRU_CACHE;
+
+typedef struct LRU_NODE_TAG
+{
+    void* key;
+    int64_t size;
+    CLDS_HASH_TABLE_ITEM* value;
+    DLIST_ENTRY node;
+    volatile_atomic int32_t state; /* LRU_NODE_STATE */
```
mattdurak, line 53: I think this should have an eviction callback/context

src/lru_cache.c
```diff
@@ -0,0 +1,505 @@
...
+        else
+        {
+            /*Codes_SRS_LRU_CACHE_13_037: [ While the current size of the cache exceeds capacity: ]*/
+
+            if (DList_IsListEmpty(&lru_cache->head))
+            {
+                /*Codes_SRS_LRU_CACHE_13_050: [ For any other errors, lru_cache_put shall return LRU_CACHE_PUT_ERROR ]*/
+                LogError("Something is wrong. The cache is empty but there is no capacity");
+                srw_lock_ll_release_exclusive(&lru_cache->srw_lock);
+                evict_callback(context, LRU_CACHE_EVICT_ERROR, NULL);
```
mattdurak, line 230: >evict_callback(context, LRU_CACHE_EVICT_ERROR, NULL); [](http://example.com/codeflow?start=16&length=53)

See my other comment, but if we change it so each item has a callback, then I guess create would need an "on_error" callback for cases like this.

src/lru_cache.c
```diff
@@ -0,0 +1,505 @@
...
+            }
+
+            /*Codes_SRS_LRU_CACHE_13_038: [ lru_cache_put shall get the least used node which is Flink of head node. ]*/
+            DLIST_ENTRY* least_used_node = lru_cache->head.Flink;
+            LRU_NODE* least_used_node_value = CONTAINING_RECORD(least_used_node, LRU_NODE, node);
+
+            if (current_size - least_used_node_value->size < 0)
+            {
+                /*Codes_SRS_LRU_CACHE_13_050: [ For any other errors, lru_cache_put shall return LRU_CACHE_PUT_ERROR ]*/
+                LogError("current_size is less 0. Failing eviction.");
```
mattdurak, line 242: >current_size is less [](http://example.com/codeflow?start=26&length=20)

really current_size minus the node size is less than 0. Should log those 2 values to understand why

src/lru_cache.c
mattdurak: You may want to keep the underflow check to be defensive but leave out its spec/test if it's not possible to test it.

src/lru_cache.c
```diff
@@ -0,0 +1,505 @@
...
+                    switch (remove_result)
+                    {
+                        case CLDS_HASH_TABLE_REMOVE_OK:
+                        {
+
+                            int32_t state = interlocked_add(&least_used_node_value->state, 0);
+                            if (state != LRU_NODE_STATE_READY)
+                            {
+                                LogWarning("invalid state(%" PRI_MU_ENUM ") for lru node. Node has already been evicted.", MU_ENUM_VALUE(LRU_NODE_STATE, state));
+                            }
```
mattdurak, line 272: If we always just lock for everything, do we even need this state anymore?

src/lru_cache.c
```diff
@@ -0,0 +1,505 @@
...
+                            LogError("item with key =%p has already been evicted.", least_used_node_value->key);
+                            (void)interlocked_add_64(&lru_cache->current_size, least_used_node_value->size);
+                            break;
+                        }
+                        case CLDS_HASH_TABLE_REMOVE_ERROR:
+                        {
+                            /*Codes_SRS_LRU_CACHE_13_050: [ For any other errors, lru_cache_put shall return LRU_CACHE_PUT_ERROR ]*/
+                            LogError("Error removing item with key =%p from hash table", least_used_node_value->key);
+                            result = LRU_CACHE_EVICT_ERROR;
+                            evict_callback(context, LRU_CACHE_EVICT_ERROR, NULL);
```
mattdurak, line 300: >LRU_CACHE_EVICT_ERROR [](http://example.com/codeflow?start=52&length=21)

likely want to break out of the loop also here

devdoc/lru_cache_design.md
```diff
@@ -31,87 +29,118 @@ typedef struct LRU_NODE_TAG
...
-    H --> I
-    I --> J
+    G --> I
+    I --> H
     J --> K
+    K --> L
+    L --> M
+    M --> P
+    P --> N
+    N --> O
```
ACKPaul, line 83: Not need if the item is already present. any how we are not going to increase the size.

src/lru_cache.c
```diff
@@ -0,0 +1,442 @@
...
+                    LRU_NODE* new_node = CLDS_HASH_TABLE_GET_VALUE(LRU_NODE, item);
+                    new_node->key = key;
+                    new_node->size = size;
+                    new_node->value = value;
+                    new_node->evict_callback = evict_callback;
+                    new_node->evict_callback_context = context;
+
+                    CLDS_HASH_TABLE_ITEM* old_item;
+                    /*Codes_SRS_LRU_CACHE_13_065: [ lru_cache_put shall update the LRU Node item in the hash table by calling clds_hash_table_set_value. ]*/
+                    CLDS_HASH_TABLE_SET_VALUE_RESULT set_value_result = clds_hash_table_set_value(lru_cache->table, hazard_pointers_thread, key, item, NULL, NULL, &old_item, NULL);
```
ACKPaul, line 330: "Why do we need to set a value here? I believe we're already reading a sealed block, and there's no chance of increasing the size. Inserting a value via the API should work fine, and all the size-related checks from the old block will become obsolete."

devdoc/lru_cache_design.md
```diff
@@ -31,87 +29,118 @@ typedef struct LRU_NODE_TAG
...
-    G --> B
-    B --> H
-    H --> I
-    I --> J
+    G --> I
+    I --> H
     J --> K
+    K --> L
+    L --> M
+    M --> P
```
ACKPaul, line 81: "I believe we're already reading a sealed block, and there's no chance of increasing the size.  I think this is not required.

ACKPaul: "If we want to consider size, there should be a flag in the LRU cache creation, giving users the option to enable or disable it."

src/lru_cache.c
ACKPaul: "If we want to consider size, there should be a flag in the LRU cache creation, giving users the option to enable or disable it."

src/lru_cache.c
```diff
@@ -0,0 +1,435 @@
...
+        (initial_bucket_size == 0) ||
+        /*Codes_SRS_LRU_CACHE_13_004: [ If clds_hazard_pointers is NULL, lru_cache_create shall fail and return NULL. ]*/
+        (clds_hazard_pointers == NULL) ||
+        /*Codes_SRS_LRU_CACHE_13_010: [ If capacity is less than or equals to 0, then lru_cache_create shall fail and return NULL. ]*/
+        (capacity <= 0) || 
+        /*Codes_SRS_LRU_CACHE_13_079: [ If on_error_callback is NULL then lru_cache_create shall fail and return NULL. ]*/
+        on_error_callback == NULL
+        )
+    {
+        LogError("Invalid arguments: COMPUTE_HASH_FUNC compute_hash=%p, KEY_COMPARE_FUNC key_compare_func=%p, uint32_t initial_bucket_size=%u, CLDS_HAZARD_POINTERS_HANDLE clds_hazard_pointers=%p, int64_t capacity=%" PRId64 "",
```
mattdurak, line 74: >int64_t capacity=%" PRId64 "", [](http://example.com/codeflow?start=196&length=30)

missing on_error parameters

src/lru_cache.c
```diff
@@ -0,0 +1,435 @@
...
+static LRU_CACHE_EVICT_RESULT evict_internal(LRU_CACHE_HANDLE lru_cache, CLDS_HAZARD_POINTERS_THREAD_HANDLE hazard_pointers_thread)
+{
+    LRU_CACHE_EVICT_RESULT result = LRU_CACHE_EVICT_OK;
+    do
+    {
+        /*Codes_SRS_LRU_CACHE_13_040: [ lru_cache_put shall acquire the lock in exclusive. ]*/
+        srw_lock_ll_acquire_exclusive(&lru_cache->srw_lock);
+
+        int64_t current_size = interlocked_add_64(&lru_cache->current_size, 0);
+        if (result != LRU_CACHE_EVICT_OK || current_size <= lru_cache->capacity)
```
mattdurak, line 170: >result != LRU_CACHE_EVICT_OK  [](http://example.com/codeflow?start=12&length=29)

nit: this could just be the "while" condition

tests/lru_cache_int/lru_cache_int.c
```diff
@@ -0,0 +1,905 @@
...
+    CLDS_HASH_TABLE_NODE_RELEASE(TEST_ITEM, item1);
+    CLDS_HASH_TABLE_NODE_RELEASE(TEST_ITEM, item2);
+    CLDS_HASH_TABLE_NODE_RELEASE(TEST_ITEM, item3);
+
+    lru_cache_destroy(lru_cache);
+    clds_hazard_pointers_destroy(hazard_pointers);
+}
+
+
+// Cyrus
```
mattdurak, line 336: ?

mattdurak (dismissed):
:shipit:

mattdurak (dismissed):
:shipit:

tests/lru_cache_ut/lru_cache_ut.c
```diff
@@ -0,0 +1,1398 @@
...
+    REGISTER_TYPE(CLDS_HASH_TABLE_INSERT_RESULT, CLDS_HASH_TABLE_INSERT_RESULT);
+    REGISTER_TYPE(CLDS_HASH_TABLE_REMOVE_RESULT, CLDS_HASH_TABLE_REMOVE_RESULT);
+
+    ASSERT_ARE_EQUAL(int, 0, real_thread_notifications_dispatcher_init());
+
+    test_clds_hazard_pointers = real_clds_hazard_pointers_create();
+    ASSERT_IS_NOT_NULL(test_clds_hazard_pointers, "Cannot create hazard pointers");
+
+    test_clds_hazard_pointers_thread_helper = real_clds_hazard_pointers_thread_helper_create(test_clds_hazard_pointers);
+    ASSERT_IS_NOT_NULL(test_clds_hazard_pointers, "Cannot create hazard pointers");
```
mattdurak, line 168: >pointers [](http://example.com/codeflow?start=72&length=8)

thread helper

tests/lru_cache_ut/lru_cache_ut.c
```diff
@@ -0,0 +1,1398 @@
...
+#include "macro_utils/macro_utils.h"
+#include "testrunnerswitcher.h"
+
+#include "real_gballoc_ll.h"
+
+#include "umock_c/umock_c.h"
+#include "umock_c/umocktypes_stdint.h"
+#include "umock_c/umock_c_negative_tests.h"
+
+#include "c_pal/interlocked.h"
```
mattdurak, line 16: >interlocked [](http://example.com/codeflow?start=16&length=11)

maybe should have a todo comment with a task number to put this back in the mocks...

mattdurak (dismissed):
:shipit:

dcristoloveanu (approved):

PR #178, opened 2023-09-13
LRU Cache Design

cyrus-jackson:
Design document for LRU Cache using CLDS_HASH_TABLE and doublylinkedlist from c_utils

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -159,6 +159,13 @@ set(clds_h_files
     ./inc/clds/mpsc_lock_free_queue.h
 )
 
+if (WIN32)
+    set(clds_h_files
+        ${clds_h_files}
+        ./inc/clds/clds_hazard_pointers_thread_helper.h # Windows only until there is a PAL for thread local storage
+    )
+endif()
+
 set(clds_c_files
     ./src/clds_hazard_pointers.c
     ./src/clds_sorted_list.c
@@ -168,6 +175,13 @@ set(clds_c_files
     ./src/clds_singly_linked_list.c
     ./src/mpsc_lock_free_queue.c
 )
+
+if (WIN32)
+    set(clds_c_files
+        ${clds_c_files}
+        ./src/clds_hazard_pointers_thread_helper.c # Windows only until there is a PAL for thread local storage
+    )
+endif()
 
 FILE(GLOB clds_md_files "devdoc/*.md")
 SOURCE_GROUP(devdoc FILES ${clds_md_files})
--- a/build/devops_gated.yml
+++ b/build/devops_gated.yml
@@ -283,13 +283,12 @@ jobs:
       configuration: RelWithDebInfo
       maximumCpuCount: true
 
-# Fails for some reason, tracked by task
-#  - script: |
-#     echo Running Hash table perf test
-#     $(Build.Repository.LocalPath)/build_x64/RelWithDebInfo/clds_hash_table_perf.exe
-#
-#    workingDirectory: 'build_x64'
-#    displayName: 'Run clds_hash_table_perf tests'
+  - script: |
+     echo Running Hash table perf test
+     $(Build.Repository.LocalPath)/build_x64/RelWithDebInfo/clds_hash_table_perf.exe
+
+    workingDirectory: 'build_x64'
+    displayName: 'Run clds_hash_table_perf tests'
 
   - script: |
      echo Running Singly linked list perf test
--- a/devdoc/clds_hash_table_requirements.md
+++ b/devdoc/clds_hash_table_requirements.md
@@ -438,6 +438,8 @@ static void on_sorted_list_skipped_seq_no(void* context, int64_t skipped_sequenc
 
 **SRS_CLDS_HASH_TABLE_01_075: [** `on_sorted_list_skipped_seq_no` called with NULL `context` shall return. **]**
 
+**SRS_CLDS_HASH_TABLE_01_113: [** If the sequence number callback passed to `clds_hash_table_create` was `NULL`, `on_sorted_list_skipped_seq_no` shall return. **]**
+
 **SRS_CLDS_HASH_TABLE_01_076: [** `on_sorted_list_skipped_seq_no` shall call the skipped sequence number callback passed to `clds_hash_table_create` and pass the `skipped_sequence_no` as `skipped_sequence_no` argument. **]**
 
 ### clds_hash_table_snapshot
--- a/devdoc/clds_hazard_pointers_thread_helper_requirements.md
+++ b/devdoc/clds_hazard_pointers_thread_helper_requirements.md
@@ -1,0 +1,70 @@
+# `clds_hazard_pointers_thread_helper` Requirements
+
+## Overview
+
+This is a helper which handles storing and retrieving thread-local hazard pointers threads.
+
+## Exposed API
+
+```c
+typedef struct CLDS_HAZARD_POINTERS_THREAD_HELPER_TAG* CLDS_HAZARD_POINTERS_THREAD_HELPER_HANDLE;
+
+MOCKABLE_FUNCTION(, CLDS_HAZARD_POINTERS_THREAD_HELPER_HANDLE, clds_hazard_pointers_thread_helper_create, CLDS_HAZARD_POINTERS_HANDLE, hazard_pointers);
+MOCKABLE_FUNCTION(, void, clds_hazard_pointers_thread_helper_destroy, CLDS_HAZARD_POINTERS_THREAD_HELPER_HANDLE, hazard_pointers_helper);
+
+MOCKABLE_FUNCTION(, CLDS_HAZARD_POINTERS_THREAD_HANDLE, clds_hazard_pointers_thread_helper_get_thread, CLDS_HAZARD_POINTERS_THREAD_HELPER_HANDLE, hazard_pointers_helper);
+```
+
+### clds_hazard_pointers_thread_helper_create
+
+```c
+MOCKABLE_FUNCTION(, CLDS_HAZARD_POINTERS_THREAD_HELPER_HANDLE, clds_hazard_pointers_thread_helper_create, CLDS_HAZARD_POINTERS_HANDLE, hazard_pointers);
+```
+
+`clds_hazard_pointers_thread_helper_create` initializes the thread-local state for using the hazard pointers and returns the handle to the helper. The `hazard_pointers` lifetime must be greater than or equal to this handle, as no reference counting is performed.
+
+**SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_42_001: [** If `hazard_pointers` is `NULL` then `clds_hazard_pointers_thread_helper_create` shall fail and return `NULL`. **]**
+
+**SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_42_002: [** `clds_hazard_pointers_thread_helper_create` shall allocate memory for the helper. **]**
+
+**SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_42_003: [** `clds_hazard_pointers_thread_helper_create` shall allocate the thread local storage slot for the hazard pointers by calling `TlsAlloc`. **]**
+
+**SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_42_004: [** `clds_hazard_pointers_thread_helper_create` shall succeed and return the helper. **]**
+
+**SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_42_005: [** If there are any errors then `clds_hazard_pointers_thread_helper_create` shall fail and return `NULL`. **]**
+
+### clds_hazard_pointers_thread_helper_destroy
+
+```c
+MOCKABLE_FUNCTION(, void, clds_hazard_pointers_thread_helper_destroy, CLDS_HAZARD_POINTERS_THREAD_HELPER_HANDLE, hazard_pointers_helper);
+```
+
+`clds_hazard_pointers_thread_helper_destroy` frees the resources from the create function.
+
+**SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_42_006: [** If `hazard_pointers_helper` is `NULL` then `clds_hazard_pointers_thread_helper_destroy` shall return. **]**
+
+**SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_42_007: [** `clds_hazard_pointers_thread_helper_destroy` shall free the thread local storage slot by calling `TlsFree`. **]**
+
+**SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_42_008: [** `clds_hazard_pointers_thread_helper_destroy` shall free the helper. **]**
+
+### clds_hazard_pointers_thread_helper_get_thread
+
+```c
+MOCKABLE_FUNCTION(, CLDS_HAZARD_POINTERS_THREAD_HANDLE, clds_hazard_pointers_thread_helper_get_thread, CLDS_HAZARD_POINTERS_THREAD_HELPER_HANDLE, hazard_pointers_helper);
+```
+
+`clds_hazard_pointers_thread_helper_get_thread` gets the `CLDS_HAZARD_POINTERS_THREAD_HANDLE` for the thread, using thread local storage to store or retrieve the handle if it has already been registered.
+
+**SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_42_009: [** If `hazard_pointers_helper` is `NULL` then `clds_hazard_pointers_thread_helper_get_thread` shall fail and return `NULL`. **]**
+
+**SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_42_010: [** `clds_hazard_pointers_thread_helper_get_thread` shall get the thread local handle by calling `TlsGetValue`. **]**
+
+**SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_42_011: [** If no thread local handle exists then: **]**
+
+ - **SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_42_012: [** `clds_hazard_pointers_thread_helper_get_thread` shall create one by calling `clds_hazard_pointers_register_thread`. **]**
+
+ - **SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_42_013: [** `clds_hazard_pointers_thread_helper_get_thread` shall store the new handle by calling `TlsSetValue`. **]**
+
+**SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_42_014: [** `clds_hazard_pointers_thread_helper_get_thread` shall return the thread local handle. **]**
+
+**SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_42_015: [** If there are any errors then `clds_hazard_pointers_thread_helper_get_thread` shall fail and return `NULL`. **]**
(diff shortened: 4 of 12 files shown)


devdoc/lru_cache_design.md
```diff
@@ -0,0 +1,118 @@
...
+
+`lru_cache` is a module that implements a Least Recently Used (LRU) cache using `clds_hash_table` and `doublylinkedlist` from `c_util`.
+
+The module provides the following functionality:
+
+- Inserting items in the cache.
+- Getting an item in the hash table by its key.
+- Evicting the least recently used item and auto-eviction of items when capacity is full.
+
+All operations can be concurrent with other operations of the same or different kinds.
```
dcristoloveanu, line 12: Nice callout

devdoc/lru_cache_design.md
```diff
@@ -0,0 +1,118 @@
...
+
+typedef struct LRU_NODE_TAG
+{
+    void* key;
+    int64_t size;
+    DLIST_ENTRY node;
+} LRU_NODE;
+```
+
+LRU Cache uses 2 `clds_hash_tables`, a `srw_lock`, and a `doubly_linked_list`.
```
dcristoloveanu, line 37: >2 `clds_hash_tables` [](http://example.com/codeflow?start=15&length=20)

Funny comment: what's better than one hash table? 2 hashtables!

devdoc/lru_cache_design.md
```diff
@@ -0,0 +1,118 @@
...
+    F --> G
+    E --> G
+    G --> H
+    H --> I
+    I --> J
+```
+
+### Eviction
+
+To evict the least recently used item from the cache, the key from the tail of the `doubly_linked_list` is removed in both the tables (i.e., `link_table` and `item_table`), and the node is truncated from the `doubly_linked_list`.
```
dcristoloveanu, line 73: >`link_table` and `item_table`), and the node is truncated from the `doubly_linked_list`. [](http://example.com/codeflow?start=141&length=88)

There seems to be 3 operations needed.
Do we really need 3?

I am thinking of the following design (and maybe it doe snot work, but I shall brain dump it here):

We have a hash table - we want this to look up the blocks/kvps.
Each of these entries stored in the hash table has a dlist entry.
we have a dlist that gives the order.

I understand we need for put:
- craft entry
- insert in hash table
- lock list
- insert in list at head (flink)
- unlock list

For find:
- find in has table

For evict:
- lock list
- remove from tail (blink)
- unlock list
- remove from hash table the item

I don't see why we need the 3rd structure? Seems to complicate things. But maybe I am missing something.

dcristoloveanu (changes requested):
:clock1:

devdoc/lru_cache_design.md
```diff
@@ -0,0 +1,118 @@
...
+- Evicting the least recently used item and auto-eviction of items when capacity is full.
+
+All operations can be concurrent with other operations of the same or different kinds.
+
+```c
+typedef struct LRU_CACHE_TAG
+{
+    CLDS_HAZARD_POINTERS_HANDLE clds_hazard_pointers;
+    CLDS_HAZARD_POINTERS_THREAD_HANDLE hazard_pointers_thread;
+    CLDS_HASH_TABLE_HANDLE item_table;
```
ACKPaul, line 19: I think we do not need two hash Table. One Hash Table is good enough for LRU cache.

devdoc/lru_cache_design.md
ACKPaul: One question: What is the rationale for requiring an explicit eviction function? In my view, if a table or list is full when inserting data, it should automatically handle eviction without the need for explicit instructions.

ACKPaul: An LRU cache should have both get(key) and put(key, value) operations to manage its contents efficiently based on the principle of retaining the most recently used items and evicting the least recently used ones when necessary. Please correct me if my understanding is not correct.

devdoc/lru_cache_design.md
```diff
@@ -0,0 +1,118 @@
+
+# LRU Cache Design
+
+`lru_cache` is a module that implements a Least Recently Used (LRU) cache using `clds_hash_table` and `doublylinkedlist` from `c_util`.
```
avranju, line 4: >`doublylinkedlist` from `c_util` [](http://example.com/codeflow?start=102&length=32)

Given that this cache lives in clds repo, is it OK for clds to take a dependency on c_util? As in, c_util doesn't depend on clds does it?

devdoc/lru_cache_design.md
```diff
@@ -0,0 +1,118 @@
...
+    participant LinkHashTable as "Linked Hash Table"
+    participant LRUList as "Doubly Linked List"
+    Cache ->> MainHashTable: Find the item in the main hash table (key)
+    MainHashTable -->> Cache: Found the item (hash_table_item)
+    Cache ->> LinkHashTable: Find the item in the linked hash table (key)
+    LinkHashTable -->> Cache: Found the item (link_item)
+    Cache ->> LRUList: Extract LRU_NODE from link_item
+    LRUList ->> LRUList: Check the item's position in the list
+    LRUList -->> Cache: Current item position
+    LRUList ->> LRUList: Acquire an exclusive lock
```
avranju, line 114: >LRUList ->> LRUList: Acquire an exclusive lock [](http://example.com/codeflow?start=4&length=46)

Good thing Dan set us straight and moved this cache into the client because this lock means that all offload reads are going to be serialized. This may not be as much of a problem on the client side as it is likely to be on the service. I am not sure how EH uses the client, but I doubt that they do a lot of concurrent reads on the same client instance. Or even if they did, that the concurrency is so great that this serialization is likely to become a bottleneck.

devdoc/lru_cache_design.md
```diff
@@ -0,0 +1,118 @@
...
+    D -->|Yes| E
+    D -->|No| F
+    F --> G
+    E --> G
+    G --> H
+    H --> I
+    I --> J
+```
+
+### Eviction
```
avranju, line 71: >Eviction [](http://example.com/codeflow?start=4&length=8)

We should probably make provision for having a callback invoked whenever stuff gets evicted. This would allow us to have metrics in higher layers to track eviction (which would allow us to tune capacity).

devdoc/lru_cache_design.md
mattdurak: Yes that is fine, see "/doc/wiki/dev/submodule dependencies.md" in the EBS repo for the dependency layout.

devdoc/lru_cache_design.md
```diff
@@ -0,0 +1,118 @@
...
+3. `lock` is used to exclusively lock the `doubly_linked_list` while changing the order or removing the node from the list.
+
+### Inserting items into the cache
+
+The `lru_cache_put` function is used to insert or update an item in the cache. If the item already exists in the cache, it is removed and reinserted to maintain the LRU order. If the cache is full, it performs eviction by removing the least recently used item until there is enough space for the new item.
+
+```mermaid
+graph TD
+    A[Input: lru_cache, key, value, size]
+    B[Acquire Lock]
```
mattdurak, line 49: > B[Acquire Lock] [](http://example.com/codeflow?start=3&length=16)

I have to wonder, if this is in CLDS (lock-less data structures), why are we wrapping the lock-less hash table with a lock :) ?

If we truly need the lock, would it be better to just have a non-threadsafe hashtable that we use because we are locking anyway?

mattdurak: Although Dan's comment below makes sense about ONLY locking the linked list and using the hash table outside of the lock. So maybe you are going to address this in a later iteration.

devdoc/lru_cache_design.md
mattdurak: Erm *cough cough*, Elastic Log chiming in here. We do many concurrent reads today.

There is a PBI to improve this so we would end up only doing multiple concurrent reads for a given request to different blocks (which should be in different partitions and thus different clients).

Maybe it's fine because the fix we need to make anyway, but just calling it out :)

devdoc/lru_cache_design.md
```diff
@@ -0,0 +1,122 @@
...
+    void* key;
+    int64_t size;
+    CLDS_HASH_TABLE_ITEM* value;
+    DLIST_ENTRY node;
+} LRU_NODE;
+
+typedef void(*LRU_CACHE_EVICT_CALLBACK_FUNC)(void* context, LRU_CACHE_EVICT_RESULT cache_evict_status);
+```
+
+LRU Cache uses `clds_hash_tables`, a `srw_lock`, and a `doubly_linked_list`.
```
mattdurak, line 39: >clds_hash_tables [](http://example.com/codeflow?start=16&length=16)

likely should be singular now

devdoc/lru_cache_design.md
```diff
@@ -0,0 +1,122 @@
...
+
+1.  The `table` stores the key and value `(void* key, CLDS_HASH_TABLE_ITEM* value)`. The `table` stores `DLIST_ENTRY` and `size` by crafting it in the form of `LRU_NODE` and storing this as a value in the form of `CLDS_HASH_TABLE_ITEM`.
```
mattdurak, line 41: >table` stores the key and value `(void* key, CLDS_HASH_TABLE_ITEM* value)`. The `table` stores `DLIST_ENTRY` [](http://example.com/codeflow?start=9&length=108)

This says the table stores two different things. Likely can be reworded in the second sentence like "the value stores the `DLIST_ENTRY`..." Or Just completely rephrase that second sentence as "The `CLDS_HASH_TABLE_ITEM` contains an `LRU_NODE` which has a `DLIST_ENTRY` and `size`."? Maybe needs to say something about the user content somewhere though.

devdoc/lru_cache_design.md
mattdurak: FYI, there is a `clds_singly_linked_list` (used internally by the hash table I think). Not sure if you really need the doubly linked list or if that could work.

devdoc/lru_cache_design.md
```diff
@@ -0,0 +1,122 @@
...
+    C --> D
+    D -->|Yes| E
+    D -->|No| F
+    F --> G
+    E --> G
+    G --> H
+    H --> I
+    I --> J
+    J --> K
+```
```
mattdurak, line 73: Can't we limit the locking a bit here to just be around the list?

```
    A --> C
    C --> D
    D -->|Yes| E
    D -->|No| F
    F --> G
    E --> G
    G --> B
    B --> H
    H --> I
    I --> J
    J --> K
```

devdoc/lru_cache_design.md
mattdurak: I'll even throw out a crazy idea:

`clds_singly_linked_list` already works for insert/delete. The missing piece is you want to evict the "tail", which means you need to walk the list to find it if it's not doubly-linked.

Well, what if you accept that eviction could be "slow" (walk the list) and we push it on another thread?

Then you could do the eviction by calling `clds_singly_linked_list_find` with a callback that just counts nodes until it reaches the eviction size and stores them in its `item_compare_context`.

Now you have a collection of nodes that should be evicted and you can do the normal removal.

devdoc/lru_cache_design.md
```diff
@@ -0,0 +1,122 @@
...
+    DoublyLinkedList -->> Cache: Item removed from the list
+    Cache ->> DoublyLinkedList: Release the exclusive lock
+    DoublyLinkedList -->> Cache: Released the lock successfully
+    Cache ->> Cache: Update the current size
+    Cache ->> Cache: Call the callback provided with the eviction status
+    Note left of Cache: Eviction Logic` 
+```
+### Getting Items from the Cache
+
+This operation retrieves items from the cache and rearranges the order in the `doubly_linked_list` by moving the found item to its head. It ensures that recently accessed items are placed at the front of the list to maintain the LRU order.
```
mattdurak, line 100: >rearranges the order in the `doubly_linked_list` [](http://example.com/codeflow?start=50&length=48)

I wonder if it would be possible to use clds_sorted_list for this? Or some modification of it. If the "item_key" was some always incrementing volatile number that is assigned, then the sorted list key compare was looking at that value, then maybe you can get a lock-less list of ordered items to evict.

In fact, eviction could maybe just be calling `clds_sorted_list_delete_key` for the older values.

Not sure if it's worth looking at this now or just putting it as a note in the improvements section.

mattdurak (approved):
:shipit:

devdoc/lru_cache_design.md
avranju, line 39: >clds_hash_tables [](http://example.com/codeflow?start=16&length=16)

Nit: `clds_hash_table`. Maybe:

_LRU Cache uses a `clds_hash_table`_

devdoc/lru_cache_design.md
```diff
@@ -0,0 +1,122 @@
...
+    E --> G
+    G --> H
+    H --> I
+    I --> J
+    J --> K
+```
+
+### Eviction Logic
+
+To evict the least recently used item from the cache, the key from the tail of the `doubly_linked_list` is removed from the `table`, and the node is truncated from the `doubly_linked_list`. When an item gets evicted, the provided callback is triggered to report the eviction status.
```
avranju, line 77: >truncated [](http://example.com/codeflow?start=149&length=9)

Removed?

devdoc/lru_cache_design.md
avranju, line 41: >The `table` stores `DLIST_ENTRY` and `size` by crafting it in the form of `LRU_NODE` and storing this as a value in the form of `CLDS_HASH_TABLE_ITEM`. [](http://example.com/codeflow?start=85&length=151)

Consider saying something like:

The `table` contains `LRU_NODE` instances as the value corresponding to the supplied keys. The `LRU_NODE` contains a `DLIST_ENTRY` which points to a node in the linked list. Do we store a pointer to the `LRU_NODE` in the linked list as well?

devdoc/lru_cache_design.md
```diff
@@ -0,0 +1,122 @@
...
+
+```mermaid
+sequenceDiagram
```
avranju, line 80: Should this logic run in a loop till cache usage drops below desired usage?

devdoc/lru_cache_design.md
```diff
@@ -0,0 +1,122 @@
...
+    LRUList ->> LRUList: Acquire an exclusive lock
+    LRUList ->> LRUList: Move the item to the front if needed
+    LRUList ->> LRUList: Release the exclusive lock
+    Cache -->> Cache: Return the hash_table_item from LRU_NODE` 
+```
+
+
+### Scope for Improvements
+
+One area of improvement lies in the management of the `doubly_linked_list`, which is currently protected by a lock. To further optimize concurrent access to the cache, a lock-free `doubly_linked_list` can be used and remove `srw_lock` in its entirety.
```
avranju, line 121: >a lock-free `doubly_linked_list` [](http://example.com/codeflow?start=168&length=32)

Another alternative could be having the eviction performed from a background thread (as discussed on the chat). Might be useful to capture the idea here so we can evaluate it later if need be.

avranju (approved):
:shipit:

devdoc/lru_cache_design.md
```diff
@@ -0,0 +1,127 @@
...
+This operation retrieves items from the cache and rearranges the order in the `doubly_linked_list` by moving the found item to its head. It ensures that recently accessed items are placed at the front of the list to maintain the LRU order.
+
+```mermaid
+sequenceDiagram
+    participant Cache as "LRU Cache"
+    participant HashTable as "Hash Table (Cache)"
+    participant LRUList as "Doubly Linked List"
+    Cache ->> HashTable: Find the item in the hash table (key)
+    HashTable -->> Cache: Found the item (hash_table_item)
+    Cache ->> LRUList: Extract LRU_NODE from the item
```
ACKPaul, line 109: To prevent race conditions and ensure data consistency, acquire a lock before reading shared data.

devdoc/lru_cache_design.md
```diff
@@ -0,0 +1,127 @@
...
+
+The `lru_cache_put` function is used to insert or update an item in the cache. If the item already exists in the cache, it is removed and reinserted in the cache and moved to head position of `doubly_linked_list` to maintain the LRU order. If the cache is full, it performs eviction by removing the least recently used item until there is enough space for the new item.
+
+```mermaid
+graph TD
+    A[Input: lru_cache, key, value, size]
+    B[Acquire Lock]
+    C[Find item in cache]
+    D[If item exists]
+    E[Remove item from cache]
```
ACKPaul, line 54: Couple of points: 

1. If an item already exists in the cache, instead of removing it and adding it again, it's more efficient to update its content after finding it. However, is there a specific reason why we can't do this with the clds_hash_table?

2. When an item is not present in the cache, it's advisable to first check if the current cache size (current_size) is equal to the capacity (capacity) before initiating the eviction logic. Eviction should only be considered when the cache is at its maximum capacity.

devdoc/lru_cache_design.md
```diff
@@ -0,0 +1,127 @@
...
+    Cache -->> Cache: Return the hash_table_item from LRU_NODE` 
+```
+
+
+### Scope for Improvements
+
+- One area of improvement lies in the management of the `doubly_linked_list`, which is currently protected by a lock. To further optimize concurrent access to the cache, a lock-free `doubly_linked_list` can be used and remove `srw_lock` in its entirety. 
+
+
+- Another potential enhancement involves moving `doubly_linked_list` updates to a background thread. Instead of immediately updating the list within cache operations like `get` we could use a queue for deferred processing. This approach enhances concurrency but may introduce race conditions between eviction and `doubly_linked_list` updates as the cache nears capacity. This might occasionally lead to extra waits during `put` operations, especially when the cache is nearly full. (See this [comment](https://github.com/Azure/clds/pull/178#discussion_r1326092733) for more context)
```
mattdurak, line 124: >This might occasionally lead to extra waits during `put` operations [](http://example.com/codeflow?start=371&length=67)

I guess that depends on whether the LRU cache **must never exceed the capacity**, or if it's acceptable to temporarily exceed (so puts can go ahead and the eviction will eventually catch up).

devdoc/lru_cache_design.md
ACKPaul: For 2: Prior to invoking the "evict" function, we can explicitly highlight or specify the necessary details in the diagram?

devdoc/lru_cache_design.md
ACKPaul: Should not we make `current_size` atomic variable here, for the same reason.

devdoc/lru_cache_design.md
ACKPaul: LRUList ->> LRUList: Acquire an exclusive lock should be before the ==> LRUList ->> LRUList: Check the item's position in the list ?

devdoc/lru_cache_design.md
ACKPaul: Step 2: You can proceed with the insertion, and if the capacity exceeds its limit, then remove the oldest entry.

What I would have implemented in c++ (it is not thread safe), a toy example: 
```
class LRUCache {
    unordered_map<int,list<pair<int,int>>::iterator> umap; 
    list<pair<int,int>> que;
    int capacity; 
    
public:
    LRUCache(int capacity) {     
        this->capacity = capacity ; 
    }
    
    int get(int key) {
        auto it = umap.find(key);         
        if(it != umap.end()) {
            
            int val = it->second->second;  
            
            movetoFront(it);           
                      
            return val;             
        }
        
        return -1; 
        
    }
    
    void put(int key, int value) {
        
         auto it = umap.find(key); 
        
         if(it != umap.end()){
             it->second->second = value; 
             movetoFront(it);
         }else {
             
             umap[key] = que.insert(que.begin(),{key,value}); 
             
             if(capacity < que.size()){
                 auto li = que.back(); 
                 umap.erase(li.first); 
                 que.pop_back();                                         
             }                        
         }        
    }
    
    void movetoFront( unordered_map<int,list<pair<int,int>>::iterator>::iterator & it) {        
        int key = it->second->first; 
        int val = it->second->second;         
        que.erase(it->second);
        it->second = que.insert(que.begin(), {key,val});              
    }
};
```

dcristoloveanu (approved):
:shipit:

PR #179, opened 2023-09-13
Add thread local hazard pointers helper

mattdurak:
Small helper for using thread local storage with hazard pointers. Note that this relies on Windows Tls API's currently (but could use a PAL version in the future)

### DIFF
--- a/src/clds_hazard_pointers_thread_helper.c
+++ b/src/clds_hazard_pointers_thread_helper.c
@@ -1,0 +1,131 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+
+#include "macro_utils/macro_utils.h"
+
+#include "windows.h" // Thread Local Storage doesn't have a PAL yet
+
+#include "c_logging/logger.h"
+
+#include "c_pal/gballoc_hl.h"
+#include "c_pal/gballoc_hl_redirect.h"
+
+#include "clds/clds_hazard_pointers.h"
+
+#include "clds/clds_hazard_pointers_thread_helper.h"
+
+typedef struct CLDS_HAZARD_POINTERS_THREAD_HELPER_TAG
+{
+    CLDS_HAZARD_POINTERS_HANDLE hazard_pointers;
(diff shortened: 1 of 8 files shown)


src/clds_hazard_pointers_thread_helper.c
```diff
@@ -0,0 +1,131 @@
...
+#include "c_pal/gballoc_hl_redirect.h"
+
+#include "clds/clds_hazard_pointers.h"
+
+#include "clds/clds_hazard_pointers_thread_helper.h"
+
+typedef struct CLDS_HAZARD_POINTERS_THREAD_HELPER_TAG
+{
+    CLDS_HAZARD_POINTERS_HANDLE hazard_pointers;
+    DWORD tls_slot;
```
dcristoloveanu, line 19: >DWORD tls_slot; [](http://example.com/codeflow?start=4&length=15)

This sounds like it would have to go in the PAL at some point...

tests/CMakeLists.txt
```diff
@@ -6,6 +6,9 @@ add_subdirectory(reals)
 #unittests
 if(${run_unittests})
     build_test_folder(reals_ut)
+    if(WIN32)
+        build_test_folder(clds_hazard_pointers_thread_helper_ut) # Windows only until there is a PAL for thread local storage
```
dcristoloveanu, line 10: > there is a PAL for thr [](http://example.com/codeflow?start=85&length=23)

haha, was thinking the same when looking at the other file.

dcristoloveanu (approved):
:shipit:

src/clds_hazard_pointers_thread_helper.c
mattdurak: Yup, I had also added a backlog item to EBS for that, but wanted to just pull this out to CLDS for the time being

tests/CMakeLists.txt
mattdurak: :)

PR #180, opened 2023-09-14
Fix and re-enable clds_hash_table_perf

dcristoloveanu:
Add a NULL check in the skipped_seq_no_cb in order to not crash in the setup that we have in the perf test.

### DIFF
--- a/tests/clds_hash_table_ut/clds_hash_table_ut.c
+++ b/tests/clds_hash_table_ut/clds_hash_table_ut.c
@@ -175,6 +175,30 @@ TEST_FUNCTION_CLEANUP(method_cleanup)
 {
 }
 
+typedef struct CLDS_HASH_TABLE_TEST_CONTEXT_TAG
+{
+    CLDS_HAZARD_POINTERS_HANDLE hazard_pointers;
+    CLDS_HAZARD_POINTERS_THREAD_HANDLE hazard_pointers_thread;
+    volatile_atomic int64_t start_seq_no;
+} CLDS_HASH_TABLE_TEST_CONTEXT;
+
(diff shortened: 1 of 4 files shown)


avranju (dismissed):
LGTM

tests/clds_hash_table_ut/clds_hash_table_ut.c
```diff
@@ -3104,6 +3104,40 @@ TEST_FUNCTION(on_sorted_list_skipped_seq_no_calls_the_hash_table_skipped_seq_no)
     clds_hazard_pointers_destroy(hazard_pointers);
 }
 
+/* Tests_SRS_CLDS_HASH_TABLE_01_113: [ If the sequence number callback passed to clds_hash_table_create was NULL, on_sorted_list_skipped_seq_no shall return. ]*/
+TEST_FUNCTION(on_sorted_list_skipped_seq_no_with_NULL_hash_table_skipped_seq_no_returns)
+{
+    // arrange
+    CLDS_HAZARD_POINTERS_HANDLE hazard_pointers = clds_hazard_pointers_create();
```
ACKPaul, line 3111: It would be helpful to centralize this code into a static function for reuse across various Test cases (above two).

ACKPaul (approved):

tests/clds_hash_table_ut/clds_hash_table_ut.c
dcristoloveanu: Adding a test context for this.

dcristoloveanu: Did some extraction into a context. Saved some hundred or so lines of code I think. Not much ROI because many tests have somewhat different setup, but there was a little ROI nevertheless.

PR #188, opened 2023-10-07
Add test for degradation with threads

dcristoloveanu:
- Add a test that reproduces the problem of perf degradation with number of threads being used

### DIFF
--- a/build/devops_gated.yml
+++ b/build/devops_gated.yml
@@ -311,6 +311,13 @@ jobs:
     workingDirectory: 'build_x64'
     displayName: 'Run lock_free_set_perf tests'
 
+  - task: CmdLine@1
+    displayName: 'Run ctest'
+    inputs:
+      filename: ctest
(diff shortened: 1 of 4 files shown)


build/devops_gated.yml
```diff
@@ -311,6 +311,13 @@ jobs:
     workingDirectory: 'build_x64'
     displayName: 'Run lock_free_set_perf tests'
 
+  - task: CmdLine@1
+    displayName: 'Run ctest'
```
mattdurak, line 315: >un ctest' [](http://example.com/codeflow?start=19&length=9)

Should this say "Run ctests - Perf" or "Run perf tests"?

mattdurak (approved):
:shipit:

build/devops_gated.yml
dcristoloveanu: Done.

mattdurak (approved):
:shipit:

PR #198, opened 2023-10-22
Add TCALL_DISPATCHER specs and header

dcristoloveanu:


devdoc/tcall_dispatcher_requirements.md
```diff
@@ -0,0 +1,166 @@
...
+int TCALL_DISPATCHER_DISPATCH_CALL(T)(TCALL_DISPATCHER(T) tcall_dispatcher, ...)
+```
+
+`TCALL_DISPATCHER_DISPATCH_CALL(T)` calls each registered call target with the parameters passed in `...`.
+
+If `tcall_dispatcher` is `NULL` then `TCALL_DISPATCHER_DISPATCH_CALL(T)` shall fail and return a non-zero value.
+
+Otherwise, `TCALL_DISPATCHER_DISPATCH_CALL(T)` shall acquire the lock in shared mode.
+
+For each call target that was registered, `TCALL_DISPATCHER_DISPATCH_CALL(T)` shall call  the `function_to_call` with the associated context and the parameters in `...`.
```
anporumb, line 162: So this call is done under a lock being held and that's ok?

How about making a copy of the list of "function_to_call" (under the lock), release the lock, and then use that list to call? Just an idea.

devdoc/tcall_dispatcher_requirements.md
```diff
@@ -0,0 +1,166 @@
+# `tcall_dispatcher` requirements
+
+## Overview
+
+`TCALL_DISPATCHER` is a module that implement dispatching calls to a number of target call receivers.
+
+`TCALL_DISPATCHER` is thread safe (registering/unregistering call targets and dispatching calls are thread safe).
```
anporumb, line 7: >thread safe [](http://example.com/codeflow?start=100&length=11)

should mention reentrancy, can any of these "target call receivers" themselves for example... call unregister?

devdoc/tcall_dispatcher_requirements.md
```diff
@@ -0,0 +1,166 @@
...
+
+`TCALL_DISPATCHER_CREATE(T)` shall call `DList_InitializeListHead` to initialize the doubly linked list that holds the target call registrations.
+
+`TCALL_DISPATCHER_CREATE(T)` shall succeed and return a non-`NULL` value.
+
+If there are any failures then `TCALL_DISPATCHER_CREATE(T)` shall fail and return `NULL`.
+
+### TCALL_DISPATCHER_REGISTER_TARGET(T)
+```c
+TCALL_DISPATCHER_TARGET_HANDLE(T) TCALL_DISPATCHER_REGISTER_TARGET(T)(TCALL_DISPATCHER(T) tcall_dispatcher, void* function_to_call, void* call_context)
```
anporumb, line 109: > void* [](http://example.com/codeflow?start=107&length=6)

one would assume that the type of this function can be typedef'd at preprocessor time, as all arguments are knows.

Just saying it would be nice to have an incomplete type instead of "void*" (what matches everything).

devdoc/tcall_dispatcher_requirements.md
```diff
@@ -0,0 +1,166 @@
...
+It introduces the APIs (as MOCKABLE_FUNCTIONS) that can be called for a `TCALL_DISPATCHER`.
+
+It introduces also the target call function type based on the arguments in `...`.
+
+It also introduces the type `TCALL_DISPATCHER_TARGET_HANDLE(T)`, which represents a handle to a call target.
+
+`...` is a list of argtype argname pairs, example usage being:
+
+```c
+TCALL_DISPATCHER_TYPE_DECLARE(FOO, int,  x, char*, y);
```
anporumb, line 80: >* [](http://example.com/codeflow?start=48&length=1)

I am not sure this (`*`) would work nicely. Maybe only with string literals. We will see I guess.

anporumb, line 80: >  [](http://example.com/codeflow?start=39&length=1)

extra space

devdoc/tcall_dispatcher_requirements.md
```diff
@@ -0,0 +1,166 @@
...
+### TCALL_DISPATCHER_TYPE_DECLARE(T, ...)
+```c
+#define TCALL_DISPATCHER_TYPE_DECLARE(T, ...)
+```
+
+`TCALL_DISPATCHER_TYPE_DECLARE(T)` is a macro to be used in a header declaration.
+
+It introduces the APIs (as MOCKABLE_FUNCTIONS) that can be called for a `TCALL_DISPATCHER`.
+
+It introduces also the target call function type based on the arguments in `...`.
```
anporumb, line 73: >target call function type [](http://example.com/codeflow?start=23&length=25)

aaaaah, is this the type in line 109 that is void* at the moment?

devdoc/tcall_dispatcher_requirements.md
```diff
@@ -0,0 +1,166 @@
...
+
+If `tcall_dispatcher` is `NULL` then `TCALL_DISPATCHER_DISPATCH_CALL(T)` shall fail and return a non-zero value.
+
+Otherwise, `TCALL_DISPATCHER_DISPATCH_CALL(T)` shall acquire the lock in shared mode.
+
+For each call target that was registered, `TCALL_DISPATCHER_DISPATCH_CALL(T)` shall call  the `function_to_call` with the associated context and the parameters in `...`.
+
+`TCALL_DISPATCHER_DISPATCH_CALL(T)` shall release the lock.
+
+`TCALL_DISPATCHER_DISPATCH_CALL(T)` shall succeed and return 0.
```
anporumb, line 166: >0 [](http://example.com/codeflow?start=61&length=1)

Frankly, this module should be part of c-util. It is used in clds, but there's no reason for it to stay in clds. Maybe future task?

anporumb, line 166: >return [](http://example.com/codeflow?start=54&length=6)

Overall, this gotten me all giggly inside. Well done!

anporumb (approved):
:shipit:

devdoc/tcall_dispatcher_requirements.md
dcristoloveanu: Yes, yes , updated. You can see how the first iteration of the thinking was with void* :-) And then I said, but meh!

devdoc/tcall_dispatcher_requirements.md
dcristoloveanu: I think it would work fine because we are not planning to hold on to the data in this module, should be just passed to the targets.

devdoc/tcall_dispatcher_requirements.md
dcristoloveanu: I hate this too. I am absolutely happy to do this, it requires a TCALL_DISPATCHER_CTX (that I even started on), but then I said I should put it off so I do not get squirrel'd.
I added tasks to add this TCALL_DISPATCHER_CTX (THANDLE based context). It will require more changes, that is why it will be done as another task, since I want to focus on the main fix now.

dcristoloveanu: [https://msazure.visualstudio.com/One/_sprints/backlog/Azure%20Messaging%20Store/One/Custom/AzureMessaging/Germanium/GeM1?workitem=25560457](https://msazure.visualstudio.com/One/_sprints/backlog/Azure%20Messaging%20Store/One/Custom/AzureMessaging/Germanium/GeM1?workitem=25560457)

devdoc/tcall_dispatcher_requirements.md
dcristoloveanu: :-)

dcristoloveanu: Yes, plan is to move it to c-util, but for fiddling purposes I want to keep it in clds until it is used once.

devdoc/tcall_dispatcher_requirements.md
dcristoloveanu: Yes, added it's true form here and removed the void*

devdoc/tcall_dispatcher_requirements.md
dcristoloveanu: Removed

devdoc/tcall_dispatcher_requirements.md
dcristoloveanu: Mentioned that re-entrancy is not supported atm.
It does require removing the call of functions under lock - which I will work on, all in due time!

anporumb (approved):
:shipit:

PR #202, opened 2023-10-28
Specs and header for tqueue

dcristoloveanu:
These are specs for tqueue, a basic queue backed by an array of a fixes size. This can be later improved to be growable, but at the moment I need just the basic version, so making it easier to review.
Has a push
Has a pop and a pop if (conditional pop with a callback condition function).


devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,209 @@
...
+
+When popping, the tail value modulo queue size is used as array index where the element is read from.
+
+A state is associated with each array entry.
+
+The possible states for an array entry are:
+
+- `NOT_USED` - No data exists at the entry.
+- `PUSHING` - Data is currently being written at the entry due to a push. Data is not safe to be popped yet.
+- `USED` - Data exists in the enty can can be popped.
```
anporumb, line 39: >enty [](http://example.com/codeflow?start=30&length=4)

entry

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,209 @@
...
+#define TQUEUE_TYPE_DECLARE(T)
+```
+
+`TQUEUE_TYPE_DECLARE(T)` is a macro to be used in a header declaration.
+
+It introduces the APIs (as MOCKABLE_FUNCTIONS) that can be called for a `TQUEUE`.
+
+
+```c
+TQUEUE_TYPE_DECLARE(int32_t);
```
anporumb, line 83: >int32_t [](http://example.com/codeflow?start=20&length=7)

what is this int32_t doing here?

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,209 @@
...
+```
+
+### TQUEUE_CREATE(T)
+```c
+static TQUEUE(T) TQUEUE_CREATE(T)(uint32_t queue_size);
+```
+
+`TQUEUE_CREATE(T)` creates a new `TQUEUE(T)`.
+
+`TQUEUE_CREATE(T)` shall call `THANDLE_MALLOC_WITH_EXTRA_SIZE` with `NULL` as dispose function and the extra size set to `queue_size` * sizeof(T).
```
anporumb, line 104: >THANDLE_MALLOC_WITH_EXTRA_SIZE [](http://example.com/codeflow?start=31&length=30)

don't we have a malloc for thandle with flex?? Let me look that up...

anporumb: ok, maybe there isn't immediately available... 

Looks like THANDLE should get one!

anporumb: And looking at the code, looks like the intention WAS there, just not exposed..

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,209 @@
...
+
+`TQUEUE_CREATE(T)` shall initialize the head and tail of the list with 0.
+
+`TQUEUE_CREATE(T)` shall initialize the state for each entry in the array used for the queue with `NOT_USED`.
```
anporumb, line 108: >for each entry in the array [](http://example.com/codeflow?start=46&length=27)

shouldn't just initialize the entry at the index 0 suffice?

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,209 @@
...
+### TQUEUE_PUSH(T)
+```c
+int TQUEUE_PUSH(T)(TQUEUE(T) tqueue, const T item)
+```
+
+`TQUEUE_PUSH(T)` pushes an item in the queue.
+
+If `tqueue` is `NULL` then `TQUEUE_PUSH(T)` shall fail and return a non-zero value.
+
+`TQUEUE_PUSH(T)` shall execute the following actions until it is either able to push the item in the queue or it fails:
```
anporumb, line 123: >is either able to push the item in the queue or it fails: [](http://example.com/codeflow?start=62&length=57)

is timeout a case of failure?

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,209 @@
...
+
+- `TQUEUE_PUSH(T)` shall obtain the current head and tail of the queue.
```
anporumb, line 125: >and [](http://example.com/codeflow?start=49&length=3)

Of course the professor leaves as an exercise to the students the exact way in which the imperative in line 7 is achieved AND 2 values are read "together".

Not saying it is impossible, but would be nice to have some better defined magic around here.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,209 @@
...
+
+- If the queue is full, `TQUEUE_PUSH(T)` shall fail and return a non-zero value.
```
anporumb, line 127: >- If the queue is full, [](http://example.com/codeflow?start=0&length=23)

one would assume that queue full condition is head==tail and element at head is used. How does the code do this verification?

Or we going to do in line 104 queue_size+1?

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,209 @@
...
+
+- If the state of the array entry corresponding to the head is not `NOT_USED`, `TQUEUE_PUSH(T)` shall try again.
+
+- If the state of the array entry corresponding to the head is `NOT_USED`:
+
+  - `TQUEUE_PUSH(T)` shall set the state to `PUSHING`.
+
+  - `TQUEUE_PUSH(T)` shall increment the head value.
+
+  - `TQUEUE_PUSH(T)` shall copy the value of `item` into the array entry value that whose state was changed to `PUSHING`.
```
anporumb, line 137: >copy [](http://example.com/codeflow?start=27&length=4)

copy?? Can we avoid like the plagues the copy and maybe make something better? I would be perfectly fine passing some helper functions that take a T* and return a T* (as a "fake" copy) which would just return the initial T* and just inc_ref it.


Parameter in line 116 is also const T.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,209 @@
...
+
+- If the queue is full, `TQUEUE_PUSH(T)` shall fail and return a non-zero value.
+
+- If the state of the array entry corresponding to the head is not `NOT_USED`, `TQUEUE_PUSH(T)` shall try again.
+
+- If the state of the array entry corresponding to the head is `NOT_USED`:
+
+  - `TQUEUE_PUSH(T)` shall set the state to `PUSHING`.
+
+  - `TQUEUE_PUSH(T)` shall increment the head value.
```
anporumb, line 135: >`TQUEUE_PUSH(T)` shall increment the head value. [](http://example.com/codeflow?start=4&length=48)

Assume 2 threads (T0, T1), assume Q is empty.

line 125 will stamp for both T0 and T1 T0(head)=0, T1(head)=0.

Line 129 will observe for both T0 and T1 the fact that the item is NOT_USED, so both will proceed.

Line 133 will change the state of the item to PUSHING (seems both threads will do just that).

Line 135 will end up with head=2 for the queue.

Then line 137 will copy for both threads, unclear if over the same location or different locations (pending clarifications)

etc.

Conclusion:

I don't think line 133 can be done just like that and if there are some interlocked that is supposed to be used, then let's explicitly state the interlocked op and what it does in human understandable format.

Here's example:

`TQUEUE_PUSH(T)` shall increment the head value **if it didn't change meanwhile** by using interlocked compare exchange.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,209 @@
...
+The possible states for an array entry are:
+
+- `NOT_USED` - No data exists at the entry.
+- `PUSHING` - Data is currently being written at the entry due to a push. Data is not safe to be popped yet.
+- `USED` - Data exists in the enty can can be popped.
+- `POPPING` - Data is being popped and should not be popped by any other thread.
+
+Note: plans exist to extend this queue into a growable queue.
+
+## Exposed API
```
anporumb, line 44: >Exposed [](http://example.com/codeflow?start=3&length=7)

missing some APIs (like .... pop/push/create)

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,209 @@
...
+
+  - `TQUEUE_PUSH(T)` shall copy the value of `item` into the array entry value that whose state was changed to `PUSHING`.
+
+  - `TQUEUE_PUSH(T)` shall set the state to `USED`.
+
+  - `TQUEUE_PUSH(T)` shall succeed and return 0.
+
+### TQUEUE_POP(T)
+```c
+TQUEUE_POP_RESULT TQUEUE_POP(T)(TQUEUE(T) tqueue, T* item)
```
anporumb, line 145: >TQUEUE_POP_RESULT [](http://example.com/codeflow?start=0&length=17)

should be listed in the header someplace with its values.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,209 @@
...
+
+- If the state of the array entry corresponding to the tail is not `USED`, `TQUEUE_POP(T)` shall try again.
+
+- If the state of the array entry corresponding to the tail is `USED`:
+
+  - `TQUEUE_POP(T)` shall set the state to `POPPING`.
+
+  - `TQUEUE_POP(T)` shall increment the tail value.
+
+  - `TQUEUE_POP(T)` shall copy array entry value that whose state was changed to `POPPING` to `item`.
```
anporumb, line 166: >copy array entry [](http://example.com/codeflow?start=26&length=16)

how does it even copy? memset? What happens if T is some flex type? 

... I really don't like "copy" in this module... let's avoid it.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,209 @@
...
+
+  - If `condition_function` returns `false`, `TQUEUE_POP_IF(T)` shall change the state back to `USED` and return `TQUEUE_POP_IF_DOES_NOT_MATCH`.
+
+  - Otherwise, `TQUEUE_POP_IF(T)` shall increment the tail value.
+
+  - `TQUEUE_POP_IF(T)` shall copy array entry value that whose state was changed to `POPPING` to `item`.
+
+  - `TQUEUE_POP_IF(T)` shall set the state to `NOT_USED`.
+
+  - `TQUEUE_POP_IF(T)` shall succeed and return `TQUEUE_POP_IF_OK`.
```
anporumb, line 209: >TQUEUE_POP_IF_OK [](http://example.com/codeflow?start=49&length=16)

So here are closing remarks for this module.

What kind of environment would this queue be used in? Answer sort of lies in line 7, where threads are mentioned (and all the APIS seem to take threading into account).

Now... if I assume that there are 3 pushers and 3 poppers for this queue, there's no synchronization between them.

By that I mean, pusher A pushes items 1, 2 and 3 to the queue. Popper X, Y and Z each get 1 item: X gets item "1", Y gets item "2" and Z gets item "3". 

So "as seen from the queue" they are ordered. However, from the poppers point of view, by the time when the poppers get to consume the items, they might just be consumed in the 2, 3, 1 order.

The only guarantee this queue has is that a popper miiiiight see the items pushes from the same pusher in their order, caveat it might not see all the items (because another popper stole some of them).

Soo... where is this queue useful? What's the environment where it can be used? What else is around it?

anporumb: Maybe I am taking here on poetry. Maybe the ordering capabilties of the queue induced errors in the reviewer and maaaaaybe all this module wants to do is dispatch from "n" pushers to "m" poppers a number of "x" items in whatever order.

Anyway, this is all very interesting.

src/tqueue.c
```diff
@@ -0,0 +1,13 @@
...
+#include <stdlib.h>
+
+#include "macro_utils/macro_utils.h"
+
+#include "c_logging/logger.h"
+
+#include "c_pal/gballoc_hl.h"
+#include "c_pal/gballoc_hl_redirect.h"
+
+#include "clds/tqueue.h"
```
anporumb, line 13: module shoul dbe part of c_util,

anporumb, line 13: >tqueue [](http://example.com/codeflow?start=15&length=6)

and will be waiting because 
1) copy should be avoided, any way we can be smart about not using memcpy/memmove etc?
2) if copy is avoided... then interlocked (note: there's no mention of interlocked, but I assume it would be used) can be avoided... 
3) if order is not essential to the environment, then is this a mere dispatcher and not a queue (because queues are ordered, did you ever try to cut in the line... LOL).
4) am curious to see where this goes.

anporumb (changes requested):
:clock1:

devdoc/tqueue_requirements.md
dcristoloveanu: No, we do not have flex, we have extra size. I believe that is because this was coded before the flex was introduced. malloc_flex was introduced to encapsulate the multiplications of malloc because we needed to clear the security issue. And whoever did that seems like it did not do the TMALLOC_FLEX. Likely because it made sense to not bulk all changes together.

So, this will get a THANDLE_MALLOC_FLEX, here is task:
[https://msazure.visualstudio.com/One/_sprints/taskboard/Azure%20Messaging%20Store/One/Custom/AzureMessaging/Germanium/GeM1?workitem=25626237](https://msazure.visualstudio.com/One/_sprints/taskboard/Azure%20Messaging%20Store/One/Custom/AzureMessaging/Germanium/GeM1?workitem=25626237)

devdoc/tqueue_requirements.md
dcristoloveanu: Fixed.

devdoc/tqueue_requirements.md
dcristoloveanu: Added the ones specific to the queue. Now let's see if you will want next to add the THANDLE ones ....

devdoc/tqueue_requirements.md
dcristoloveanu: It is an example usage, added the "Example usage:" text.

devdoc/tqueue_requirements.md
dcristoloveanu: Nope, at the moment we do not have timeout. When time comes we'll add it.

devdoc/tqueue_requirements.md
dcristoloveanu: Clarified that we're using interlocked operations and the fact we're reading the values in 2 steps.

devdoc/tqueue_requirements.md
dcristoloveanu: Added clarification (funny that I added that while fixing another comment, before reading this actual comment about it)

devdoc/tqueue_requirements.md
dcristoloveanu: That is exactly what I had in mind too (actually at some point 2 months ago I started with a prototype exactly along those lines).

I think we should actually have both.
For convenience of coding if we have an array of int32_t or just one field, it convolutes the code and so much more the tests also to have callbacks/helpers to set/get the values.

But in cases where there's quite a bit of data in the structures it absolutely makes sense to have some helpers filling in the data directly in the array entry and reading it from there.

Added flavors of the APIs with this

devdoc/tqueue_requirements.md
dcristoloveanu: Haha, another comment that I addressed before reading the actual comment (because of the earlier comment saying we should specify how we touch these).

It says now interlocked_compare_excahnge is used.

devdoc/tqueue_requirements.md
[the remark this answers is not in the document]
dcristoloveanu: I added flavors or push/pop/pop_if with helper functions (like I said this was on my mind too, but I did not add it because current use case did not need that). But why not, let's have them.

As for locks - technically we'd want 2 locks since one of the features I'd like to have for this "queue is to not block pushers on poppers and poppers on pushers when not needed (if queue is full, pushers will be blocked on poppers of course and the other way around).

I'd rather stick to the interlocked for keep in track of head, tail and the state of each element in the queue as we have done up to now for many cases (ordered_op_writer, latency tracker in zrpc, several cases in test code, threadpool for Linux etc.).

I'd just want to generalize this and hopefully get rid of all the code instances we already have (after I am done with the current customer fix - shall not name it :-))

devdoc/tqueue_requirements.md
dcristoloveanu: Added a section detailing which useful APIs are generated (PUSH/POP/POP_IF/,,,

devdoc/tqueue_requirements.md
dcristoloveanu: Then we'd have to initialize the next entry every time we move head or tail and things get more complicated.
Should not be a big deal to init once all of them. We do that for all cases where we use this pattern already.

src/tqueue.c
dcristoloveanu: It will absolutely be once I am done with the current fix. Task exists for it in the current PBI:

[https://msazure.visualstudio.com/One/_sprints/taskboard/Azure%20Messaging%20Store/One/Custom/AzureMessaging/Germanium/GeM1?workitem=25560482](https://msazure.visualstudio.com/One/_sprints/taskboard/Azure%20Messaging%20Store/One/Custom/AzureMessaging/Germanium/GeM1?workitem=25560482)

devdoc/tqueue_requirements.md
dcristoloveanu: Technically maybe for this use case of mine, a bag would be enough.
So I could go and rename this to bag and be done.
But!
The guarantees of this MPMC queue is that:
- if there is only one consumer and one producer, order is observed by the consumer as it was produced.
- if there are multiple producers and multiple consumers, order is not guaranteed, it is best effort.

There are at least 4-5 instances where we have built this kind of queue (sometimes MPMC, sometimes SPSC, sometimes, MPSC).

One such example is in threadpool Linux - it is MPMC. We have pushers there which put items to execute in the "bag".
And we have poppers taking them out to execute them.
Now in that case we could completely ignore order and say it is a bag. On the other hand having a best effort to execute the items in the order they came in is not a bad thing.

One of the reasons we end up building these queues backed by some fixed size arrays (or sometimes not so fixed) is because we don't want to allocate memory for each item. Allocations are expensive.

I could go and have a SPSC queue backed by an array implementation and then a separate MPMC queue, but I think one implementation should be enough for now for us....

devdoc/tqueue_requirements.md
dcristoloveanu: Added 2 APIs that will allow using helpers to get/set

src/tqueue.c
dcristoloveanu: 1) should be addressed with the addition of the extra APIs with helper functions.
2) If we do want that most times producers and consumers don't hurt each other then ... we kind of need the interlocked. Again, I'd just writing this once so we don't write it every day as I see a couple of these implementations in our code.
3) Sometimes it is when we have SCSP, sometimes it is not (if the user uses multiple consumers or multiple producers then so be it, it's on the user :-)
4) Curiosity might or might not be satisfied. Let me know.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+/*to be used in a .c file*/
+#define TQUEUE_TYPE_DEFINE(T)
+```
+
+The macros expand to these useful somewhat more useful APIs:
+
+```c
+TQUEUE(T) TQUEUE_CREATE(T)(uint32_t queue_size);
+int TQUEUE_PUSH_WITH_SET_FUNC(T)(TQUEUE(T) tqueue, const T item);
```
anporumb, line 71: >SET_FUNC [](http://example.com/codeflow?start=21&length=8)

404 SET_FUNC not found.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+TQUEUE_POP_RESULT TQUEUE_POP(T)(TQUEUE(T) tqueue, T* item);
+TQUEUE_POP_IF_RESULT TQUEUE_POP_IF(T)(TQUEUE(T) tqueue, T* item, TQUEUE_DEFINE_CONDITION_FUNCTION_TYPE_NAME(T), condition_function, void*, condition_function_context);
```
anporumb, line 73: >T* item [](http://example.com/codeflow?start=56&length=7)

this sounds like an "out" argument, any chance we might have it the last one? Right not it sits between the handle and the function

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+int TQUEUE_PUSH(T)(TQUEUE(T) tqueue, const T item);
+TQUEUE_POP_RESULT TQUEUE_POP(T)(TQUEUE(T) tqueue, T* item);
```
anporumb, line 75: >TQUEUE_POP [](http://example.com/codeflow?start=18&length=10)

what's the difference between this and line 72?

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+TQUEUE_POP_IF_RESULT TQUEUE_POP_IF(T)(TQUEUE(T) tqueue, T* item, TQUEUE_DEFINE_CONDITION_FUNCTION_TYPE_NAME(T), condition_function, void*, condition_function_context);
```
anporumb, line 76: >TQUEUE_POP_IF [](http://example.com/codeflow?start=21&length=13)

what's the difference between this and line 76??

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+/*to be used in a .c file*/
+#define TQUEUE_TYPE_DEFINE(T)
+```
+
+The macros expand to these useful somewhat more useful APIs:
+
+```c
+TQUEUE(T) TQUEUE_CREATE(T)(uint32_t queue_size);
+int TQUEUE_PUSH_WITH_SET_FUNC(T)(TQUEUE(T) tqueue, const T item);
+TQUEUE_POP_RESULT TQUEUE_POP(T)(TQUEUE(T) tqueue, T* item);
```
anporumb, line 72: >TQUEUE_POP [](http://example.com/codeflow?start=18&length=10)

looks like POP is a variant of POP_IF called with a "return true" function.

Maybe provide "bool POP_ALWAYS()" function and then don't have 2 APIs.

However... left at the imagination of the original code writer.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+TQUEUE_POP_IF_RESULT TQUEUE_POP_IF(T)(TQUEUE(T) tqueue, T* item, TQUEUE_DEFINE_CONDITION_FUNCTION_TYPE_NAME(T), condition_function, void*, condition_function_context);
+```
+
+The signature of the condition function is:
+
+```c
+bool TQUEUE_DEFINE_CONDITION_FUNCTION_TYPE_NAME(T)(void* context, const T* item);
+```
+
+The signature of the set function is:
```
anporumb, line 85: >set [](http://example.com/codeflow?start=21&length=3)

Frankly, what has "set" must have "unset" too.

Case in point: if the queue contains "THANDLEs". Putting a THANDLE in the queue<===>THANDLE_INC_REF.

When the queue is destroyed... who knows how to free these thandles? So something like "unset" MUST exist someplace. 404 not found.

anporumb: I was about to mumble something about "how did the test did not catch this", but lo and behold, there's no tests available for review at this time.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+```c
+bool TQUEUE_DEFINE_SET_FUNCTION_TYPE_NAME(T)(void* context, T* item);
+```
+
+The signature of the get function is:
+
+```c
+bool TQUEUE_DEFINE_GET_FUNCTION_TYPE_NAME(T)(void* context, const T* item);
```
anporumb, line 94: >TQUEUE_DEFINE_GET_FUNCTION_TYPE_NAME [](http://example.com/codeflow?start=5&length=36)

what GET function??? where is it used? Gets what...

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+`TQUEUE_PUSH(T)` shall execute the following actions until it is either able to push the item in the queue or it fails:
+
+- `TQUEUE_PUSH(T)` shall obtain the current head queue by calling `interlocked_add_64`.
+
+- `TQUEUE_PUSH(T)` shall obtain the current tail queue by calling `interlocked_add_64`.
+
+- If the queue is full (current head >= current tail + queue size), `TQUEUE_PUSH(T)` shall fail and return a non-zero value.
+
+- If the state of the array entry corresponding to the head is not `NOT_USED`, `TQUEUE_PUSH(T)` shall try again.
```
anporumb, line 172: >`TQUEUE_PUSH(T)` shall try again. [](http://example.com/codeflow?start=79&length=33)

So if we assume this scenario with 3 very fast pushers and 3 very fast popper... 

... and for some reason the poppers are blocked (maybe waiting on network, who knows, doesn't matter)... then the pushers will 
1) fill all the slots in the queue very fast and then
2) ... clog 3 threads in a "while(1)" loop?

This is not nice. What are the chances it will happen in production when somebody is using this module and naively assumes "some" behavior... 

May I dare say in this case, retries should be surfaced to the user and the user can decide what to do.
1) retry immediately (hammering)
2) sleep exponentially and then retry
3) give up after a while
4) other creative strategy

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+- If the state of the array entry corresponding to the head is `NOT_USED`:
+
+  - `TQUEUE_PUSH(T)` shall set the state to `PUSHING` (from `NOT_USED`) by using `interlocked_compare_exchange`.
+
+  - `TQUEUE_PUSH(T)` shall increment the head value.
+
+  - `TQUEUE_PUSH(T)` shall copy the value of `item` into the array entry value that whose state was changed to `PUSHING`.
```
anporumb, line 180: >that [](http://example.com/codeflow?start=79&length=4)

extra "that"?

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+If `tqueue` is `NULL` then `TQUEUE_PUSH(T)` shall fail and return a non-zero value.
+
+`TQUEUE_PUSH(T)` shall execute the following actions until it is either able to push the item in the queue or it fails:
+
+- `TQUEUE_PUSH(T)` shall obtain the current head queue by calling `interlocked_add_64`.
+
+- `TQUEUE_PUSH(T)` shall obtain the current tail queue by calling `interlocked_add_64`.
+
+- If the queue is full (current head >= current tail + queue size), `TQUEUE_PUSH(T)` shall fail and return a non-zero value.
```
anporumb, line 170: >urrent head >= current tail + queue size [](http://example.com/codeflow?start=25&length=40)

somewhere here it would be nice to state that head, tail are forever increasing (looks like that) and all the indexes are % queue size.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+  - `TQUEUE_PUSH(T)` shall succeed and return 0.
+
+### TQUEUE_POP(T)
+```c
+TQUEUE_POP_RESULT TQUEUE_POP(T)(TQUEUE(T) tqueue, T* item)
+```
+
+`TQUEUE_POP(T)` pops an item from the queue if available.
+
+If `tqueue` is `NULL` then `TQUEUE_POP(T)` shall fail and return `TQUEUE_POP_ERROR`.
```
anporumb, line 193: >ERROR [](http://example.com/codeflow?start=77&length=5)

invalid_arg

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+  - `TQUEUE_PUSH(T)` shall copy the value of `item` into the array entry value that whose state was changed to `PUSHING`.
+
+  - `TQUEUE_PUSH(T)` shall set the state to `USED` by using `interlocked_exchange`.
+
+  - `TQUEUE_PUSH(T)` shall succeed and return 0.
+
+### TQUEUE_POP(T)
+```c
+TQUEUE_POP_RESULT TQUEUE_POP(T)(TQUEUE(T) tqueue, T* item)
```
anporumb, line 188: >TQUEUE_POP_RESULT [](http://example.com/codeflow?start=0&length=17)

still not defined in this iteration.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+`TQUEUE_POP(T)` shall execute the following actions until it is either able to pop the item from the queue or it fails:
+
+- `TQUEUE_POP(T)` shall obtain the current head queue by calling `interlocked_add_64`.
+
+- `TQUEUE_POP(T)` shall obtain the current tail queue by calling `interlocked_add_64`.
+
+- If the queue is empty (current tail == current head), `TQUEUE_POP(T)` shall fail and return `TQUEUE_POP_QUEUE_EMPTY`.
+
+- If the state of the array entry corresponding to the tail is not `USED`, `TQUEUE_POP(T)` shall try again.
```
anporumb, line 203: >`TQUEUE_POP(T)` shall try again. [](http://example.com/codeflow?start=75&length=32)

So ... if somehow this stays "pushing" for a very long time, then all the new poppers will "while(1)". 

Should be surfaced to the user this conundrum and let the user decide what he wants to do with it.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+- If the state of the array entry corresponding to the tail is `USED`:
+
+  - `TQUEUE_POP(T)` shall set the state to `POPPING` (from `USED`) by using `interlocked_compare_exchange`.
+
+  - `TQUEUE_POP(T)` shall replace the tail value with the tail value obtained earlier + 1 by using `interlocked_compare_exchange`.
```
anporumb: >interlocked_compare_exchange [](http://example.com/codeflow?start=100&length=28)

So will this be retried... (if cmpxchg fails then by virtue of line 195 it will fail.).

But that would limit the popper to effectively just 1 ... 

So I am going to assume that differnet poppers CAN come and they will not impede each other.

Maybe rewrite so it is clear

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+  - `TQUEUE_POP(T)` shall copy array entry value that whose state was changed to `POPPING` to `item`.
```
anporumb, line 211: >that [](http://example.com/codeflow?start=49&length=4)

extra 
that"?

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+### TQUEUE_POP(T)
+```c
+TQUEUE_POP_RESULT TQUEUE_POP(T)(TQUEUE(T) tqueue, T* item)
+```
+
+`TQUEUE_POP(T)` pops an item from the queue if available.
+
+If `tqueue` is `NULL` then `TQUEUE_POP(T)` shall fail and return `TQUEUE_POP_ERROR`.
+
+`TQUEUE_POP(T)` shall execute the following actions until it is either able to pop the item from the queue or it fails:
```
anporumb, line 195: >fails [](http://example.com/codeflow?start=113&length=5)

failure case is not exactly clear here.

queue empty is not a failure.

I don't see anything else that can "fail". It either copies (cough) the items and succeeds or queue is empty. 

So what are the failure cases here?

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+TQUEUE_POP_IF_RESULT TQUEUE_POP_IF(T)(TQUEUE(T) tqueue, T* item, TQUEUE_DEFINE_CONDITION_FUNCTION_TYPE_NAME(T), condition_function, void*, condition_function_context)
+```
+
+`TQUEUE_POP_IF(T)` pops an item from the queue conditionally.
+
+If `tqueue` is `NULL` then `TQUEUE_POP_IF(T)` shall fail and return `TQUEUE_POP_IF_ERROR`.
+
+`TQUEUE_POP_IF(T)` shall execute the following actions until it is either able to pop the item from the queue or it fails:
+
+- `TQUEUE_POP_IF(T)` shall obtain the current head and tail of the queue.
```
anporumb, line 228: should use the verbiage from lines 197-199

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+If `tqueue` is `NULL` then `TQUEUE_POP(T)` shall fail and return `TQUEUE_POP_ERROR`.
+
+`TQUEUE_POP(T)` shall execute the following actions until it is either able to pop the item from the queue or it fails:
+
+- `TQUEUE_POP(T)` shall obtain the current head queue by calling `interlocked_add_64`.
+
+- `TQUEUE_POP(T)` shall obtain the current tail queue by calling `interlocked_add_64`.
+
+- If the queue is empty (current tail == current head), `TQUEUE_POP(T)` shall fail and return `TQUEUE_POP_QUEUE_EMPTY`.
```
anporumb, line 201: >fail [](http://example.com/codeflow?start=78&length=4)

queue empty is not exactly a failure case. Normal behavior.

devdoc/tqueue_requirements.md
anporumb: >interlocked_compare_exchange [](http://example.com/codeflow?start=100&length=28)

I don't think you are going to like this comment of mine now.

Assume there are 3 threads in the system.

T0 is the poor thread trying to get the element (and doesn't get too mcuh CPU time allocated)

T1 and T2 are its mean brothers who are clogging at the CPU time. T1 will do push, T2 does pop.

Let's say the queue is 3 element big, and there's an element at index 0.

Having introduced these actors, what happens in this scenario?

T0 (the slow one) comes and evaluates line 201. It sees the queue as "not empty", great! Then it takes a rather LONG break.

Meanwhile T1 comes and does 2 pushes (at this moment qeueue is full)

T2 comes and does 3 pops (note: T0 is still immediately after executing line 201). At this moment the queue is empty, and the first location where data is written is index 0. 

T1 comes and does ANOTHER push (this populates index0).

T0 finally gets some CPU time and will execute 


FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFS.

Line 209 really needs to say it retries.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+- If the state of the array entry corresponding to the tail is not `USED`, `TQUEUE_POP(T)` shall try again.
+
+- If the state of the array entry corresponding to the tail is `USED`:
+
+  - `TQUEUE_POP(T)` shall set the state to `POPPING` (from `USED`) by using `interlocked_compare_exchange`.
```
anporumb, line 207: >interlocked_compare_exchange [](http://example.com/codeflow?start=77&length=28)

lines 203, 205 and 207 *must* be all the same interlocked. This one that switches from USER to POPPING.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+- `TQUEUE_PUSH(T)` shall obtain the current tail queue by calling `interlocked_add_64`.
+
+- If the queue is full (current head >= current tail + queue size), `TQUEUE_PUSH(T)` shall fail and return a non-zero value.
+
+- If the state of the array entry corresponding to the head is not `NOT_USED`, `TQUEUE_PUSH(T)` shall try again.
+
+- If the state of the array entry corresponding to the head is `NOT_USED`:
+
+  - `TQUEUE_PUSH(T)` shall set the state to `PUSHING` (from `NOT_USED`) by using `interlocked_compare_exchange`.
```
anporumb, line 176: >interlocked_compare_exchange [](http://example.com/codeflow?start=82&length=28)

same observation here, there can only be 1 cmpxchg for the state. Lines 172, 174 and 176 can be re-written to make that obvious.

devdoc/tqueue_requirements.md
anporumb, line 170: >fail [](http://example.com/codeflow?start=91&length=4)

queue full /empty, just like bank accounts are not failures, but rather facts of life.

I don't see anything that "fails" here.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+- If the state of the array entry corresponding to the tail is not `USED`, `TQUEUE_POP_IF(T)` shall try again.
+
+- If the state of the array entry corresponding to the tail is `USED`:
+
+  - `TQUEUE_POP_IF(T)` shall set the state to `POPPING` (from `USED`) by using `interlocked_compare_exchange`.
+
+  - `TQUEUE_POP_IF(T)` shall call `condition_function`, passing to it `condition_function_context` and the address of the `T` value in the array entry corresponding to the tail of the queue.
+
+  - If `condition_function` returns `false`, `TQUEUE_POP_IF(T)` shall change the state back to `USED` and return `TQUEUE_POP_IF_DOES_NOT_MATCH`.
```
anporumb, line 240: >DOES_NOT_MATCH [](http://example.com/codeflow?start=128&length=14)

There's no matching done here, actually we don't know what the user would do in the callback.

At this moment I would change this poetry and call it TQUEUE_POP_IF_USER_PREVENTED. Or similar.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+  - `TQUEUE_POP(T)` shall replace the tail value with the tail value obtained earlier + 1 by using `interlocked_compare_exchange`.
+
+  - `TQUEUE_POP(T)` shall copy array entry value that whose state was changed to `POPPING` to `item`.
+
+  - `TQUEUE_POP(T)` shall set the state to `NOT_USED` by using `interlocked_exchange`.
+
+  - `TQUEUE_POP(T)` shall succeed and return `TQUEUE_POP_OK`.
+
+### TQUEUE_POP_IF(T)
```
anporumb, line 217: >TQUEUE_POP_IF [](http://example.com/codeflow?start=4&length=13)

I feel a whole lot of copy&paste between thius function and its sister without "IF".

Can we please have just 1? It is perfectly fine to pass a NULL function or a ALWAYS_POP function provided by this module to specify that "yeah, always evaluate the condition ion line 240 as "true".

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+  - `TQUEUE_POP_IF(T)` shall copy array entry value that whose state was changed to `POPPING` to `item`.
+
+  - `TQUEUE_POP_IF(T)` shall set the state to `NOT_USED` by using `interlocked_exchange`.
+
+  - `TQUEUE_POP_IF(T)` shall succeed and return `TQUEUE_POP_IF_OK`.
+
+### TQUEUE_PUSH_WITH_SET_FUNC(T)
+```c
+int TQUEUE_PUSH_WITH_SET_FUNC(T)(TQUEUE(T) tqueue, TQUEUE_SET_FUNC(T) set_function, void* set_function_context)
```
anporumb, line 252: >TQUEUE_SET_FUNC [](http://example.com/codeflow?start=51&length=15)

I still don't know how this user supplied function looks like.

anporumb, line 252: >TQUEUE_SET_FUNC [](http://example.com/codeflow?start=51&length=15)

also same comment : can we please have just 1 "PUSH" variety that can take a NULL pointer/ ALWAYS_PUSH function.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+- If the queue is full (current head >= current tail + queue size), `TQUEUE_PUSH_WITH_SET_FUNC(T)` shall fail and return a non-zero value.
+
+- If the state of the array entry corresponding to the head is not `NOT_USED`, `TQUEUE_PUSH_WITH_SET_FUNC(T)` shall try again.
+
+- If the state of the array entry corresponding to the head is `NOT_USED`:
+
+  - `TQUEUE_PUSH_WITH_SET_FUNC(T)` shall set the state to `PUSHING` (from `NOT_USED`) by using `interlocked_compare_exchange`.
+
+  - `TQUEUE_PUSH_WITH_SET_FUNC(T)` shall increment the head value.
```
anporumb: > shall increment the head value. [](http://example.com/codeflow?start=34&length=32)

cmpxchg way of incrementing, right?

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+  - `TQUEUE_PUSH_WITH_SET_FUNC(T)` shall call `set_function` with `set_function_context` and the pointer to the array entry value that whose state was changed to `PUSHING`.
```
anporumb, line 275: >set_function [](http://example.com/codeflow?start=47&length=12)

Can this "set" function fail?

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+  - `TQUEUE_PUSH_WITH_SET_FUNC(T)` shall set the state to `USED` by using `interlocked_exchange`.
+
+  - `TQUEUE_PUSH_WITH_SET_FUNC(T)` shall succeed and return 0.
+
+### TQUEUE_POP_WITH_GET_FUNC(T)
+```c
+TQUEUE_POP_RESULT TQUEUE_POP_WITH_GET_FUNC(T)(TQUEUE(T) tqueue, T* item, TQUEUE_SET_FUNC(T) get_function, void* get_function_context)
```
anporumb, line 283: >TQUEUE_SET_FUNC [](http://example.com/codeflow?start=73&length=15)

line 283 sort of points out this should be a GET_FUNC variety. Of cource, in the absence of these function type definitions, it is hard to actually interpolate what they could possibly mean.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+  - `TQUEUE_POP_WITH_GET_FUNC(T)` shall replace the tail value with the tail value obtained earlier + 1 by using `interlocked_compare_exchange`.
+
+  - `TQUEUE_POP_WITH_GET_FUNC(T)` shall call `get_function` with `get_function_context` and the pointer to the array entry value that whose state was changed to `POPPING`.
+
+  - `TQUEUE_POP_WITH_GET_FUNC(T)` shall set the state to `NOT_USED` by using `interlocked_exchange`.
+
+  - `TQUEUE_POP_WITH_GET_FUNC(T)` shall succeed and return `TQUEUE_POP_WITH_GET_FUNC_OK`.
+
+### TQUEUE_POP_IF_WITH_GET_FUNC_WITH_GET_FUNC(T)
```
anporumb, line 312: >WITH_GET_FUNC_WITH_GET_FUNC [](http://example.com/codeflow?start=18&length=27)

So this is a functiod that if a function that delivers another function?

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+`TQUEUE_POP_IF_WITH_GET_FUNC(T)` pops an item from the queue conditionally by using a get function to access the data.
+
+If `tqueue` is `NULL` then `TQUEUE_POP_IF_WITH_GET_FUNC(T)` shall fail and return `TQUEUE_POP_IF_WITH_GET_FUNC_ERROR`.
+
+`TQUEUE_POP_IF_WITH_GET_FUNC(T)` shall execute the following actions until it is either able to pop the item from the queue or it fails:
+
+- `TQUEUE_POP_IF_WITH_GET_FUNC(T)` shall obtain the current head and tail of the queue.
+
+- If the queue is empty (current tail == current head), `TQUEUE_POP_IF_WITH_GET_FUNC(T)` shall fail and return `TQUEUE_POP_IF_WITH_GET_FUNC_QUEUE_EMPTY`.
```
anporumb, line 325: >TQUEUE_POP_IF_WITH_GET_FUNC_QUEUE_EMPTY [](http://example.com/codeflow?start=112&length=39)

Can these QUEUE_EMPTY/QEUEU_FULL return values be shared across APIs?

devdoc/tqueue_requirements.md
anporumb, line 283: >TQUEUE_SET_FUNC [](http://example.com/codeflow?start=73&length=15)

so I think I might be starting to get it here.

TQEUEU_SET_FUNC and TQUEUE_GET_FUNC have the same porototype. One copies from the user in to queue, the other copies from the queue to the user.

But essentiall ythey do the same thing: take const T* and a T* and copy from the const to the T*.

Maybe we only need 1 function prototype here.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+  - `TQUEUE_POP_WITH_GET_FUNC(T)` shall call `get_function` with `get_function_context` and the pointer to the array entry value that whose state was changed to `POPPING`.
+
+  - `TQUEUE_POP_WITH_GET_FUNC(T)` shall set the state to `NOT_USED` by using `interlocked_exchange`.
+
+  - `TQUEUE_POP_WITH_GET_FUNC(T)` shall succeed and return `TQUEUE_POP_WITH_GET_FUNC_OK`.
+
+### TQUEUE_POP_IF_WITH_GET_FUNC_WITH_GET_FUNC(T)
+```c
+TQUEUE_POP_IF_RESULT TQUEUE_POP_IF_WITH_GET_FUNC(T)(TQUEUE(T) tqueue, T* item, TQUEUE_DEFINE_CONDITION_FUNCTION_TYPE_NAME(T), condition_function, void*, condition_function_context, TQUEUE_SET_FUNC(T) get_function, void* get_function_context)
```
anporumb, line 314: >TQUEUE_POP_IF_WITH_GET_FUNC [](http://example.com/codeflow?start=21&length=27)

We should only have 1 function that takes both a "if" and a "get/set" func.

That func can be in a LL layer.

And the nicer functions can be a in HL layer.

devdoc/tqueue_requirements.md
anporumb: kk

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,343 @@
...
+
+Example usage:
+
+```c
+TQUEUE_TYPE_DEFINE(int32_t);
+```
+
+### TQUEUE_CREATE(T)
+```c
+static TQUEUE(T) TQUEUE_CREATE(T)(uint32_t queue_size);
```
anporumb, line 140: >queue_size [](http://example.com/codeflow?start=43&length=10)

This is not a nice comment, but... all sizes are either too big or too small.

What would take here to have this queue resize itself? That would be AWESOME.

devdoc/tqueue_requirements.md
anporumb: got it

devdoc/tqueue_requirements.md
anporumb, line 73: >TQUEUE_DEFINE_CONDITION_FUNCTION_TYPE_NAME [](http://example.com/codeflow?start=65&length=42)

Given that we just starting having callbacks, it should mention somewhere the reentrancy model.

Can one of these "IF"/"SET/GET" functions call other APIs of the module?

src/tqueue.c
anporumb: 1) nice
2) we can talk about this in the most hypothetical way Monday in the office.Note to self: what's the cheaper SRW lock or several interlocked?
3) maybe it has devolved into poetry. Leave it.
4) very satisfied, but I still want to see the next iteration, so "temporarily satiated".

So still waiting on different curiosities.

anporumb: and one more thing.

Assume the queue has like 3 items in it and suddenly THANDLE_DEC_REF is called and the queue gets to 0 ref.

How does it clear the "copied" items? Somewher there should be a "unset"/"destroy" function.

I fear we are going COM route (with IUnknown...). Not necessarily a bad fear.

devdoc/tqueue_requirements.md
dcristoloveanu: I decided to name these PUSH_CB, POP_CB and DISPOSE.
They have a signature mentioned now in the spec.

devdoc/tqueue_requirements.md
dcristoloveanu: I think there were significant changes to say that this comment is somewhat outdated.

devdoc/tqueue_requirements.md
dcristoloveanu: The 2 sets of functions have been removed.

devdoc/tqueue_requirements.md
dcristoloveanu: The 2 sets of functions have been removed.

devdoc/tqueue_requirements.md
dcristoloveanu: There is no more POP_IF

devdoc/tqueue_requirements.md
dcristoloveanu: Discussed in person and we ended up having 3 functions: push_cb, pop_cb, dispose.

devdoc/tqueue_requirements.md
dcristoloveanu: This sounds like an improvement that could be done at some point. Let's have first the base of it and use it...

devdoc/tqueue_requirements.md
dcristoloveanu: Fixed.

devdoc/tqueue_requirements.md
dcristoloveanu: There is now a reference about indexes being modulo.

devdoc/tqueue_requirements.md
dcristoloveanu: Sure

devdoc/tqueue_requirements.md
dcristoloveanu: Changed to say that or queue is empty.

devdoc/tqueue_requirements.md
dcristoloveanu: Removed "fail and"

devdoc/tqueue_requirements.md
dcristoloveanu: Added a task for a timeout improvement for this. I don't really need it now and we don't use it now in any of our use cases in the code (does not mean we should not have it eventually).

dcristoloveanu: [https://msazure.visualstudio.com/One/_sprints/taskboard/Azure%20Messaging%20Store/One/Custom/AzureMessaging/Germanium/GeM1?workitem=25637690](https://msazure.visualstudio.com/One/_sprints/taskboard/Azure%20Messaging%20Store/One/Custom/AzureMessaging/Germanium/GeM1?workitem=25637690)

devdoc/tqueue_requirements.md
dcristoloveanu: Added a note that ln 205 executes part of the same interlocked_compare_exchange.

devdoc/tqueue_requirements.md
dcristoloveanu: Made one function now.

devdoc/tqueue_requirements.md
dcristoloveanu: There is now a signature

devdoc/tqueue_requirements.md
dcristoloveanu: Added verbiage that it is done with the same interlocked_compare_exchange

devdoc/tqueue_requirements.md
dcristoloveanu: Added an enum so that we tell the user what happened and also removed the fail part of the spec.

devdoc/tqueue_requirements.md
dcristoloveanu: Added a return code of rejected

devdoc/tqueue_requirements.md
dcristoloveanu: API has been changed to remove duplication

devdoc/tqueue_requirements.md
dcristoloveanu: Yes, spec has been changed here

dcristoloveanu: Actually it was changed in TQUEUE_PUSH since TQUEUE_PUSH_WITH_SET_FUNC does not exist anymore.

devdoc/tqueue_requirements.md
dcristoloveanu: Right now it is void. I would say let's take this step by step. When time comes :-) we can add a return code if needed

devdoc/tqueue_requirements.md
dcristoloveanu: Signatures have been added.

devdoc/tqueue_requirements.md
dcristoloveanu: Now it is just a function that calls another function :-).

devdoc/tqueue_requirements.md
dcristoloveanu: Now we only have one API, so shared by design.

devdoc/tqueue_requirements.md
dcristoloveanu: I kept it separate: we have push_cb and pop_cb.

devdoc/tqueue_requirements.md
dcristoloveanu: I have a comment that we will do that. It has been done for at least one implementation what we have in threadpool Linux.

But as a separate item.

devdoc/tqueue_requirements.md
dcristoloveanu: We have one function now taking a pop_cb function and that one can decide if it accepts or rejects the pop.

src/tqueue.c
dcristoloveanu: A dispose_item_function has been added that can be passed to TQUEUE_CREATE

devdoc/tqueue_requirements.md
dcristoloveanu: Yes, it limits the poppers grabbing items to pop, but the actual popping can be done in parallel by multiple poppers.

devdoc/tqueue_requirements.md
dcristoloveanu: Fixed.

devdoc/tqueue_requirements.md
dcristoloveanu: Based on the latest iteration replacing the tail happens if:
- right away after setting the state to POPPING no pop function is used
- after the pop cb function has returned if result was OK. I know, I know we could have the condition function separated to minimize contention, we could even build this somewhat smarter I think, buuuuut, all in due time. ...

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,278 @@
...
+- A push callback (which is invoked by the queue as a result of a push call, allowing the user to fill in the information in T rather than performing a memory copy).
+A typical use for a queue of a `THANDLE` would be to have a `THANDLE_INITIALIZE` call performed in the push function callback.
+
+- A pop callback (which is invoked by the queue as a result of a pop call, allowing the user to extract any information from T rather than performing a memory copy).
+A typical use for a queue of a `THANDLE` would be to have a `THANDLE_MOVE` or `THANDLE_INITIALIZE_MOVE` call performed in the push function callback.
+
+- A dispose item function (which is invoked by the queue when the queue is disposed and there are still items in the queue).
+A typical use for a queue of a `THANDLE` would be to have a `THANDLE_ASSIGN` with `NULL` performed in the dispose function callback.
+
+If no push/pop/dispose callbacks functions are sepcified by the user, the queue copies the memory of T passed to `TQUEUE_PUSH`, respectively `TQUEUE_POP`.
```
anporumb, line 30: >sepcified [](http://example.com/codeflow?start=47&length=9)

are what now?

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,278 @@
...
+
+Example usage:
+
+```c
+TQUEUE_TYPE_DEFINE(int32_t);
+```
+
+### TQUEUE_CREATE(T)
+```c
+TQUEUE(T) TQUEUE_CREATE(T)(uint32_t queue_size, TQUEUE_PUSH_CB_FUNC(T) push_cb_function, TQUEUE_POP_CB_FUNC(T) pop_cb_function, TQUEUE_DISPOSE_ITEM_FUNC(T) dispose_item_function, void* dispose_function_context);
```
anporumb, line 172: >queue_size [](http://example.com/codeflow?start=36&length=10)

queue size==0 not valid, should be a spec about that.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,278 @@
...
+
+- `TQUEUE_PUSH(T)` shall obtain the current head queue by calling `interlocked_add_64`.
+
+- `TQUEUE_PUSH(T)` shall obtain the current tail queue by calling `interlocked_add_64`.
+
+- If the queue is full (current head >= current tail + queue size), `TQUEUE_PUSH(T)` shall return `TQUEUE_PUSH_QUEUE_FULL`.
+
+- If the state of the array entry corresponding to the head is not `NOT_USED`, `TQUEUE_PUSH(T)` shall try again.
+
+- Part of the same `interlocked_compare_exchange` as above, if the state of the array entry corresponding to the head is `NOT_USED`:
```
anporumb, line 223: >as above [](http://example.com/codeflow?start=50&length=8)

above doesn't have any interlocked_compare_exchange. It might have had once upon a time.

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,278 @@
...
+
+  - `TQUEUE_PUSH(T)` shall set the state to `PUSHING` (from `NOT_USED`) by using `interlocked_compare_exchange`.
+
+  - `TQUEUE_PUSH(T)` shall replace the head value with the head value obtained earlier + 1 by using `interlocked_exchange_64`.
```
anporumb, line 227: >interlocked_exchange_64 [](http://example.com/codeflow?start=101&length=23)

hmmmm.... hmmmm.... are you sure this is right here?

devdoc/tqueue_requirements.md
```diff
@@ -0,0 +1,278 @@
...
+- Push an element at the head of the queue.
+- Pop an element from the tail of the queue. Pop returns the oldest element in the queue.
+
+The module allows the user to specify 3 different callback functions:
+
+- A push callback (which is invoked by the queue as a result of a push call, allowing the user to fill in the information in T rather than performing a memory copy).
+A typical use for a queue of a `THANDLE` would be to have a `THANDLE_INITIALIZE` call performed in the push function callback.
+
+- A pop callback (which is invoked by the queue as a result of a pop call, allowing the user to extract any information from T rather than performing a memory copy).
+A typical use for a queue of a `THANDLE` would be to have a `THANDLE_MOVE` or `THANDLE_INITIALIZE_MOVE` call performed in the push function callback.
```
anporumb, line 25: I think this function does 2 things.
1) can copy
2) can filter.

Maybe would be nice to have these 2 purposes with their return values and the implemented sematic explained in detail.

devdoc/tqueue_requirements.md
anporumb, line 30: >pop [](http://example.com/codeflow?start=11&length=3)

should state that the default POP function will never filter and will always copy memory.

anporumb (approved):
:shipit:

devdoc/tqueue_requirements.md
dcristoloveanu: lol, fixed.

devdoc/tqueue_requirements.md
dcristoloveanu: Added spec, thanks!

devdoc/tqueue_requirements.md
dcristoloveanu: Added.

devdoc/tqueue_requirements.md
dcristoloveanu: 2 return values frankly says 2 functions. I think having a structure with 2 things is just asking for a horribly looking API :-). It's not like we ran out of functions.

dcristoloveanu: I'll do that likely as part of the coding. Let me get a feel how it looks.

devdoc/tqueue_requirements.md
dcristoloveanu: Added a mention of interlocked_compare_exchange in the "above" spec.

devdoc/tqueue_requirements.md
dcristoloveanu: Pretty sure :-). This should not be possible because of the state - I was imagining what I could code there and there is nothing really ...

anporumb (approved):
:shipit:

PR #204, opened 2023-11-01
Add TQUEUE tests and code

dcristoloveanu:
- Add TQUEUE unit tests and code
- Add TQUEUE int tests: simple int tests, chaos tests, chaos test with THANDLE
- Minor changes were needed to the algorithm due to chaos tests uncovering the issues (who said chaos tests ever catch anything!)
- Condition function was separated from the pop callback. Initial .md said they should go together but the mangling of concepts was: a) not nice and b) actually led to more complicated code

### DIFF
--- a/devdoc/tqueue_requirements.md
+++ b/devdoc/tqueue_requirements.md
@@ -18,11 +18,8 @@ The module provides the following functionality:
 
 The module allows the user to specify 3 different callback functions:
 
-- A push callback (which is invoked by the queue as a result of a push call, allowing the user to fill in the information in T rather than performing a memory copy).
-A typical use for a queue of a `THANDLE` would be to have a `THANDLE_INITIALIZE` call performed in the push function callback.
-
-- A pop callback (which is invoked by the queue as a result of a pop call, allowing the user to extract any information from T rather than performing a memory copy).
-A typical use for a queue of a `THANDLE` would be to have a `THANDLE_MOVE` or `THANDLE_INITIALIZE_MOVE` call performed in the push function callback.
+- A copy item callback (which is invoked by the queue as a result of a push call or a pop call, allowing the user to fill in the information in/from T rather than performing a memory copy).
+A typical use for a queue of a `THANDLE` would be to have a `THANDLE_ASSIGN` from the source to the destination parameters.
 
 - A dispose item function (which is invoked by the queue when the queue is disposed and there are still items in the queue).
 A typical use for a queue of a `THANDLE` would be to have a `THANDLE_ASSIGN` with `NULL` performed in the dispose function callback.
@@ -31,7 +28,7 @@ A typical use for a queue of a `THANDLE` would be to have a `THANDLE_ASSIGN` wit
 
 If no push/pop/dispose callbacks functions are specified by the user, the queue copies the memory of T passed to `TQUEUE_PUSH`, respectively `TQUEUE_POP`.
 
-The callbacks used for push/pop/dispose/pop condition check item are not re-entrant, they should not call `TQUEUE` APIs on the same queue.
+The callbacks used for copy item/dispose item/pop condition are not re-entrant, they should not call `TQUEUE` APIs on the same queue.
 
 Because `TQUEUE` is a kind of `THANDLE`, all of `THANDLE`'s APIs apply to `TQUEUE`. For convenience the following macros are provided out of the box with the same semantics as those of `THANDLE`'s:
 
@@ -93,37 +90,27 @@ MU_DEFINE_ENUM(TQUEUE_POP_RESULT, TQUEUE_POP_RESULT_VALUES);
 The macros expand to these useful somewhat more useful APIs:
 
 ```c
-TQUEUE(T) TQUEUE_CREATE(T)(uint32_t queue_size, TQUEUE_PUSH_CB_FUNC(T) push_cb_function, TQUEUE_POP_CB_FUNC(T) pop_cb_function, TQUEUE_DISPOSE_ITEM_FUNC(T) dispose_item_function, void* dispose_item_function_context);
-int TQUEUE_PUSH(T)(TQUEUE(T) tqueue, T* item, void* push_function_context)
-TQUEUE_POP_RESULT TQUEUE_POP(T)(TQUEUE(T) tqueue, T* item, void* pop_function_context, TQUEUE_DEFINE_CONDITION_FUNCTION_TYPE_NAME(T), condition_function, void*, condition_function_context);
+TQUEUE(T) TQUEUE_CREATE(T)(uint32_t queue_size, TQUEUE_COPY_ITEM_FUNC(T) copy_item_function, TQUEUE_DISPOSE_ITEM_FUNC(T) dispose_item_function, void* dispose_item_function_context);
+int TQUEUE_PUSH(T)(TQUEUE(T) tqueue, T* item, void* copy_function_context)
+TQUEUE_POP_RESULT TQUEUE_POP(T)(TQUEUE(T) tqueue, T* item, void* copy_function_context, TQUEUE_DEFINE_CONDITION_FUNCTION_TYPE_NAME(T), condition_function, void*, condition_function_context);
 ```
 
-The signature of the push callback function `TQUEUE_PUSH_CB_FUNC(T)` is:
+The signature of the push callback function `TQUEUE_COPY_ITEM_FUNC(T)` is:
 
 ```c
-void TQUEUE_DEFINE_PUSH_CB_FUNCTION_TYPE_NAME(T)(void* context, T* push_dst, T* push_src);
+void TQUEUE_DEFINE_COPY_ITEM_FUNCTION_TYPE_NAME(T)(void* context, T* push_dst, T* push_src);
 ```
 
-Note that `push_dst` is the pointer in the queue memory, `push_src` is the pointer to the user owned `T` being pushed in the queue.
-
-The signature of the pop callback function `TQUEUE_POP_CB_FUNC(T)` is:
-
-```c
-void TQUEUE_DEFINE_POP_CB_FUNCTION_TYPE_NAME(T)(void* context, T* pop_dst, T* pop_src);
-```
-
-Note that `pop_dst` is the user T pointer which receives the popped item, `pop_src` is the pointer to the queue memory.
-
-The signature of the dispose function `TQUEUE_DISPOSE_FUNC(T)` is:
+The signature of the dispose function `TQUEUE_DISPOSE_ITEM_FUNC(T)` is:
 
 ```c
 void TQUEUE_DEFINE_DISPOSE_FUNCTION_TYPE_NAME(T)(void* context, T* item);
 ```
 
-The signature of the dispose function `TQUEUE_CONDITION_FUNC(T)` is:
+The signature of the condition function `TQUEUE_CONDITION_FUNC(T)` is:
 
 ```c
-bool TQUEUE_DEFINE_DISPOSE_FUNCTION_TYPE_NAME(T)(void* context, T* item);
+bool TQUEUE_DEFINE_CONDITION_FUNCTION_TYPE_NAME(T)(void* context, T* item);
 ```
 
 It returns `true` if the condition is satisfied and the pop should be performed, `false` otherwise.
@@ -171,14 +158,14 @@ TQUEUE_TYPE_DEFINE(int32_t);
 
 ### TQUEUE_CREATE(T)
 ```c
-TQUEUE(T) TQUEUE_CREATE(T)(uint32_t queue_size, TQUEUE_PUSH_CB_FUNC(T) push_cb_function, TQUEUE_POP_CB_FUNC(T) pop_cb_function, TQUEUE_DISPOSE_ITEM_FUNC(T) dispose_item_function, void* dispose_item_function_context);
+TQUEUE(T) TQUEUE_CREATE(T)(uint32_t queue_size, TQUEUE_COPY_ITEM_FUNC(T) copy_item_function, TQUEUE_DISPOSE_ITEM_FUNC(T) dispose_item_function, void* dispose_item_function_context);
 ```
 
 `TQUEUE_CREATE(T)` creates a new `TQUEUE(T)`.
 
 **SRS_TQUEUE_01_001: [** If `queue_size` is 0, `TQUEUE_CREATE(T)` shall fail and return `NULL`. **]**
 
-**SRS_TQUEUE_01_002: [** If any of `push_cb_function`, `pop_cb_function` and `dispose_item_function` is `NULL` and at least one of them is not `NULL`, `TQUEUE_CREATE(T)` shall fail and return `NULL`. **]**
+**SRS_TQUEUE_01_002: [** If any of `copy_item_function` and `dispose_item_function` is `NULL` and at least one of them is not `NULL`, `TQUEUE_CREATE(T)` shall fail and return `NULL`. **]**
 
 **SRS_TQUEUE_01_003: [** `TQUEUE_CREATE(T)` shall call `THANDLE_MALLOC_FLEX` with `TQUEUE_DISPOSE_FUNC(T)` as dispose function, `nmemb` set to `queue_size` and `size` set to `sizeof(T)`. **]**
 
@@ -207,7 +194,7 @@ void TQUEUE_DISPOSE_FUNC(T)(TQUEUE(T) tqueue);
 
 ### TQUEUE_PUSH(T)
 ```c
-TQUEUE_PUSH_RESULT TQUEUE_PUSH(T)(TQUEUE(T) tqueue, T* item, void* push_cb_function_context)
+TQUEUE_PUSH_RESULT TQUEUE_PUSH(T)(TQUEUE(T) tqueue, T* item, void* copy_item_function_context)
 ```
 
 `TQUEUE_PUSH(T)` pushes an item in the queue.
@@ -232,9 +219,9 @@ TQUEUE_PUSH_RESULT TQUEUE_PUSH(T)(TQUEUE(T) tqueue, T* item, void* push_cb_funct
 
   - **SRS_TQUEUE_01_023: [** If the state of the array entry corresponding to the head is not `NOT_USED`, `TQUEUE_PUSH(T)` shall try changing the state again. **]**
 
-- **SRS_TQUEUE_01_019: [** If no `push_cb_function` was specified in `TQUEUE_CREATE(T)`, `TQUEUE_PUSH(T)` shall copy the value of `item` into the array entry value whose state was changed to `PUSHING`. **]**
+- **SRS_TQUEUE_01_019: [** If no `copy_item_function` was specified in `TQUEUE_CREATE(T)`, `TQUEUE_PUSH(T)` shall copy the value of `item` into the array entry value whose state was changed to `PUSHING`. **]**
 
-- **SRS_TQUEUE_01_024: [** If a `push_cb_function` was specified in `TQUEUE_CREATE(T)`, `TQUEUE_PUSH(T)` shall call the `push_cb_function` with `push_cb_function_context` as `context`, a pointer to the array entry value whose state was changed to `PUSHING` as `push_dst` and `item` as `push_src`. **]**
+- **SRS_TQUEUE_01_024: [** If a `copy_item_function` was specified in `TQUEUE_CREATE(T)`, `TQUEUE_PUSH(T)` shall call the `copy_item_function` with `copy_item_function_context` as `context`, a pointer to the array entry value whose state was changed to `PUSHING` as `push_dst` and `item` as `push_src`. **]**
 
 - **SRS_TQUEUE_01_020: [** `TQUEUE_PUSH(T)` shall set the state to `USED` by using `interlocked_exchange`. **]**
 
@@ -275,12 +262,14 @@ TQUEUE_POP_RESULT TQUEUE_POP(T)(TQUEUE(T) tqueue, T* item, void* pop_function_co
 
     - **SRS_TQUEUE_01_044: [** If incrementing the tail by using `interlocked_compare_exchange_64` does not succeed, `TQUEUE_POP(T)` shall revert the state of the array entry to `USED` and retry. **]**
 
-  - **SRS_TQUEUE_01_032: [** If a `pop_cb_function` was not specified in `TQUEUE_CREATE(T)`: **]**
+  - **SRS_TQUEUE_01_032: [** If a `copy_item_function` was not specified in `TQUEUE_CREATE(T)`: **]**
   
     - **SRS_TQUEUE_01_033: [** `TQUEUE_POP(T)` shall copy array entry value whose state was changed to `POPPING` to `item`. **]**
 
-  - **SRS_TQUEUE_01_037: [** If a `pop_cb_function` was specified in `TQUEUE_CREATE(T)`: **]**
+  - **SRS_TQUEUE_01_037: [** If `copy_item_function` and `sispose_item_function` were specified in `TQUEUE_CREATE(T)`: **]**
   
-    - **SRS_TQUEUE_01_038: [** `TQUEUE_POP(T)` shall call `pop_cb_function` with `pop_cb_function_context` as `context`, the array entry value whose state was changed to `POPPING` to `item` as `pop_src` and `item` as `pop_dst`. **]**
+    - **SRS_TQUEUE_01_038: [** `TQUEUE_POP(T)` shall call `copy_item_function` with `copy_item_function_context` as `context`, the array entry value whose state was changed to `POPPING` to `item` as `pop_src` and `item` as `pop_dst`. **]**
+
+    - **SRS_TQUEUE_01_045: [** `TQUEUE_POP(T)` shall call `dispose_item_function` with `dispose_item_function_context` as `context` and the array entry value whose state was changed to `POPPING` as `item`. **]**
 
   - **SRS_TQUEUE_01_034: [** `TQUEUE_POP(T)` shall set the state to `NOT_USED` by using `interlocked_exchange`, succeed and return `TQUEUE_POP_OK`. **]**
--- a/devdoc/tqueue_requirements.md
+++ b/devdoc/tqueue_requirements.md
@@ -19,7 +19,7 @@ The module provides the following functionality:
 The module allows the user to specify 3 different callback functions:
 
 - A copy item callback (which is invoked by the queue as a result of a push call or a pop call, allowing the user to fill in the information in/from T rather than performing a memory copy).
-A typical use for a queue of a `THANDLE` would be to have a `THANDLE_ASSIGN` from the source to the destination parameters.
+A typical use for a queue of a `THANDLE` would be to have a `THANDLE_INITIALIZE`, initializing the `dst` with the `src`.
 
 - A dispose item function (which is invoked by the queue when the queue is disposed and there are still items in the queue).
 A typical use for a queue of a `THANDLE` would be to have a `THANDLE_ASSIGN` with `NULL` performed in the dispose function callback.
--- a/devdoc/tqueue_requirements.md
+++ b/devdoc/tqueue_requirements.md
@@ -211,13 +211,13 @@ TQUEUE_PUSH_RESULT TQUEUE_PUSH(T)(TQUEUE(T) tqueue, T* item, void* copy_item_fun
 
 - **SRS_TQUEUE_01_022: [** If the queue is full (current head >= current tail + queue size), `TQUEUE_PUSH(T)` shall return `TQUEUE_PUSH_QUEUE_FULL`. **]**
 
-- **SRS_TQUEUE_01_018: [** Using `interlocked_compare_exchange_64`, `TQUEUE_PUSH(T)` shall replace the head value with the head value obtained earlier + 1. **]**
-
-- **SRS_TQUEUE_01_043: [** If the queue head has changed, `TQUEUE_PUSH(T)` shall try again. **]**
-
 - **SRS_TQUEUE_01_017: [** Using `interlocked_compare_exchange`, `TQUEUE_PUSH(T)` shall change the head array entry state to `PUSHING` (from `NOT_USED`). **]**
 
-  - **SRS_TQUEUE_01_023: [** If the state of the array entry corresponding to the head is not `NOT_USED`, `TQUEUE_PUSH(T)` shall try changing the state again. **]**
+  - **SRS_TQUEUE_01_023: [** If the state of the array entry corresponding to the head is not `NOT_USED`, `TQUEUE_PUSH(T)` shall retry the whole push. **]**
+
+- **SRS_TQUEUE_01_018: [** Using `interlocked_compare_exchange_64`, `TQUEUE_PUSH(T)` shall replace the head value with the head value obtained earlier + 1. **]**
+
+- **SRS_TQUEUE_01_043: [** If the queue head has changed, `TQUEUE_PUSH(T)` shall set the state back to `NOT_USED` and retry the push. **]**
 
 - **SRS_TQUEUE_01_019: [** If no `copy_item_function` was specified in `TQUEUE_CREATE(T)`, `TQUEUE_PUSH(T)` shall copy the value of `item` into the array entry value whose state was changed to `PUSHING`. **]**
 
--- a/devdoc/tqueue_requirements.md
+++ b/devdoc/tqueue_requirements.md
@@ -197,7 +197,7 @@ void TQUEUE_DISPOSE_FUNC(T)(TQUEUE(T) tqueue);
 TQUEUE_PUSH_RESULT TQUEUE_PUSH(T)(TQUEUE(T) tqueue, T* item, void* copy_item_function_context)
 ```
 
-`TQUEUE_PUSH(T)` pushes an item in the queue.
+`TQUEUE_PUSH(T)` pushes an item in the queue at the head.
 
 **SRS_TQUEUE_01_012: [** If `tqueue` is `NULL` then `TQUEUE_PUSH(T)` shall fail and return `TQUEUE_PUSH_INVALID_ARG`. **]**
 
@@ -232,7 +232,7 @@ TQUEUE_PUSH_RESULT TQUEUE_PUSH(T)(TQUEUE(T) tqueue, T* item, void* copy_item_fun
 TQUEUE_POP_RESULT TQUEUE_POP(T)(TQUEUE(T) tqueue, T* item, void* pop_function_context, TQUEUE_CONDITION_FUNC(T), condition_function, void*, condition_function_context);
 ```
 
-`TQUEUE_POP(T)` pops an item from the queue if available.
+`TQUEUE_POP(T)` pops an item from the queue (from the tail) if available.
 
 **SRS_TQUEUE_01_025: [** If `tqueue` is `NULL` then `TQUEUE_POP(T)` shall fail and return `TQUEUE_POP_INVALID_ARG`. **]**
 
@@ -244,7 +244,7 @@ TQUEUE_POP_RESULT TQUEUE_POP(T)(TQUEUE(T) tqueue, T* item, void* pop_function_co
 
 - **SRS_TQUEUE_01_029: [** `TQUEUE_POP(T)` shall obtain the current tail queue by calling `interlocked_add_64`. **]**
 
-- **SRS_TQUEUE_01_035: [** If the queue is empty (current tail == current head), `TQUEUE_POP(T)` shall return `TQUEUE_POP_QUEUE_EMPTY`. **]**
+- **SRS_TQUEUE_01_035: [** If the queue is empty (current tail >= current head), `TQUEUE_POP(T)` shall return `TQUEUE_POP_QUEUE_EMPTY`. **]**
 
 - **SRS_TQUEUE_01_030: [** Using `interlocked_compare_exchange`, `TQUEUE_PUSH(T)` shall set the tail array entry state to `POPPING` (from `USED`). **]**
 
--- a/devdoc/tqueue_requirements.md
+++ b/devdoc/tqueue_requirements.md
@@ -27,11 +27,11 @@ A typical use for a queue of a `THANDLE` would be to have a `THANDLE_MOVE` or `T
 - A dispose item function (which is invoked by the queue when the queue is disposed and there are still items in the queue).
 A typical use for a queue of a `THANDLE` would be to have a `THANDLE_ASSIGN` with `NULL` performed in the dispose function callback.
 
+- A condition function used for allowing the pop of an item. If the called condition function returns `true`, the pop is performed. If it returns `false`, the pop is not performed.
+
 If no push/pop/dispose callbacks functions are specified by the user, the queue copies the memory of T passed to `TQUEUE_PUSH`, respectively `TQUEUE_POP`.
 
-The callbacks used for push/pop/dispose item are not re-entrant, they should not call `TQUEUE` APIs on the same queue.
-
-When using a pop callback function the queue also allows the user to reject a pop (thus effectively implementing a conditional pop).
+The callbacks used for push/pop/dispose/pop condition check item are not re-entrant, they should not call `TQUEUE` APIs on the same queue.
 
 Because `TQUEUE` is a kind of `THANDLE`, all of `THANDLE`'s APIs apply to `TQUEUE`. For convenience the following macros are provided out of the box with the same semantics as those of `THANDLE`'s:
 
@@ -80,12 +80,6 @@ MU_DEFINE_ENUM(TQUEUE_PUSH_RESULT, TQUEUE_PUSH_RESULT_VALUES);
 
 MU_DEFINE_ENUM(TQUEUE_POP_RESULT, TQUEUE_POP_RESULT_VALUES);
 
-#define TQUEUE_POP_CB_FUNCTION_RESULT_VALUES \
-    TQUEUE_POP_CB_FUNCTION_OK, \
-    TQUEUE_POP_CB_FUNCTION_POP_REJECTED
-
-MU_DEFINE_ENUM(TQUEUE_POP_CB_FUNCTION_RESULT, TQUEUE_POP_CB_FUNCTION_RESULT_VALUES);
-
 /*to be used as the type of handle*/
 #define TQUEUE(T)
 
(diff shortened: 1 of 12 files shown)


jebrando (dismissed):
:shipit:

devdoc/tqueue_requirements.md
```diff
@@ -99,33 +93,41 @@ MU_DEFINE_ENUM(TQUEUE_POP_CB_FUNCTION_RESULT, TQUEUE_POP_CB_FUNCTION_RESULT_VALU
...
 Note that `pop_dst` is the user T pointer which receives the popped item, `pop_src` is the pointer to the queue memory.
 
 The signature of the dispose function `TQUEUE_DISPOSE_FUNC(T)` is:
 
 ```c
-void TQUEUE_DEFINE_DISPOSE_FUNCTION_TYPE_NAME(T)(void* context, const T* item);
+void TQUEUE_DEFINE_DISPOSE_FUNCTION_TYPE_NAME(T)(void* context, T* item);
+```
+
+The signature of the dispose function `TQUEUE_CONDITION_FUNC(T)` is:
```
anporumb, line 123: >dispose [](http://example.com/codeflow?start=21&length=7)

of the who?

devdoc/tqueue_requirements.md
```diff
@@ -99,33 +93,41 @@ MU_DEFINE_ENUM(TQUEUE_POP_CB_FUNCTION_RESULT, TQUEUE_POP_CB_FUNCTION_RESULT_VALU
...
+void TQUEUE_DEFINE_PUSH_CB_FUNCTION_TYPE_NAME(T)(void* context, T* push_dst, T* push_src);
 ```
 
 Note that `push_dst` is the pointer in the queue memory, `push_src` is the pointer to the user owned `T` being pushed in the queue.
 
 The signature of the pop callback function `TQUEUE_POP_CB_FUNC(T)` is:
 
 ```c
-TQUEUE_POP_CB_FUNCTION_RESULT TQUEUE_DEFINE_POP_CB_FUNCTION_TYPE_NAME(T)(void* context, T* pop_dst, const T* pop_src);
+void TQUEUE_DEFINE_POP_CB_FUNCTION_TYPE_NAME(T)(void* context, T* pop_dst, T* pop_src);
```
anporumb, line 112: >T [](http://example.com/codeflow?start=63&length=1)

should be const, as it is source for something.

devdoc/tqueue_requirements.md
```diff
@@ -99,33 +93,41 @@ MU_DEFINE_ENUM(TQUEUE_POP_CB_FUNCTION_RESULT, TQUEUE_POP_CB_FUNCTION_RESULT_VALU
...
+TQUEUE(T) TQUEUE_CREATE(T)(uint32_t queue_size, TQUEUE_PUSH_CB_FUNC(T) push_cb_function, TQUEUE_POP_CB_FUNC(T) pop_cb_function, TQUEUE_DISPOSE_ITEM_FUNC(T) dispose_item_function, void* dispose_item_function_context);
+int TQUEUE_PUSH(T)(TQUEUE(T) tqueue, T* item, void* push_function_context)
 TQUEUE_POP_RESULT TQUEUE_POP(T)(TQUEUE(T) tqueue, T* item, void* pop_function_context, TQUEUE_DEFINE_CONDITION_FUNCTION_TYPE_NAME(T), condition_function, void*, condition_function_context);
 ```
 
 The signature of the push callback function `TQUEUE_PUSH_CB_FUNC(T)` is:
 
 ```c
-void TQUEUE_DEFINE_PUSH_CB_FUNCTION_TYPE_NAME(T)(void* context, T* push_dst, const T* push_src);
+void TQUEUE_DEFINE_PUSH_CB_FUNCTION_TYPE_NAME(T)(void* context, T* push_dst, T* push_src);
```
anporumb, line 104: >* [](http://example.com/codeflow?start=78&length=1)

should be const as it is the source

devdoc/tqueue_requirements.md
```diff
@@ -99,33 +93,41 @@ MU_DEFINE_ENUM(TQUEUE_POP_CB_FUNCTION_RESULT, TQUEUE_POP_CB_FUNCTION_RESULT_VALU
...
 
 ```c
-void TQUEUE_DEFINE_DISPOSE_FUNCTION_TYPE_NAME(T)(void* context, const T* item);
+void TQUEUE_DEFINE_DISPOSE_FUNCTION_TYPE_NAME(T)(void* context, T* item);
+```
+
+The signature of the dispose function `TQUEUE_CONDITION_FUNC(T)` is:
+
+```c
+bool TQUEUE_DEFINE_DISPOSE_FUNCTION_TYPE_NAME(T)(void* context, T* item);
```
anporumb, line 126: >bool [](http://example.com/codeflow?start=0&length=4)

bool return value for a dispose?

devdoc/tqueue_requirements.md
```diff
@@ -99,33 +93,41 @@ MU_DEFINE_ENUM(TQUEUE_POP_CB_FUNCTION_RESULT, TQUEUE_POP_CB_FUNCTION_RESULT_VALU
...
 ```
 
+It returns `true` if the condition is satisfied and the pop should be performed, `false` otherwise.
+
```
anporumb, line 130: Are these actually related to TQUEUE_DEFINE_CONDITION_FUNCTION_TYPE_NAME??

devdoc/tqueue_requirements.md
```diff
@@ -99,33 +93,41 @@ MU_DEFINE_ENUM(TQUEUE_POP_CB_FUNCTION_RESULT, TQUEUE_POP_CB_FUNCTION_RESULT_VALU
 The macros expand to these useful somewhat more useful APIs:
 
 ```c
-TQUEUE(T) TQUEUE_CREATE(T)(uint32_t queue_size, TQUEUE_PUSH_CB_FUNC(T) push_function, TQUEUE_POP_CB_FUNC(T) pop_function, TQUEUE_DISPOSE_ITEM_FUNC(T) dispose_item_function, void* dispose_function_context);
-int TQUEUE_PUSH(T)(TQUEUE(T) tqueue, const T item, void* push_function_context)
+TQUEUE(T) TQUEUE_CREATE(T)(uint32_t queue_size, TQUEUE_PUSH_CB_FUNC(T) push_cb_function, TQUEUE_POP_CB_FUNC(T) pop_cb_function, TQUEUE_DISPOSE_ITEM_FUNC(T) dispose_item_function, void* dispose_item_function_context);
```
anporumb, line 96: >CB [](http://example.com/codeflow?start=60&length=2)

I will note that the poetry here is not really callback-y, it is rather "copy". as in PUSH_COPY_FUNC, cause it wants to copy/clone/inc_ref/whatever.

Same for the other "callbacks".

anporumb: Perfectly fine to not be done in this PR, but if so entertained, in a separate "rename" PR.

devdoc/tqueue_requirements.md
```diff
@@ -195,86 +197,90 @@ void TQUEUE_DISPOSE_FUNC(T)(TQUEUE(T) tqueue);
...
-For each item in the queue, `dispose_item_function` shall be called with `dispose_function_context` and a pointer to the array entry value (T*).
+**SRS_TQUEUE_01_011: [** For each item in the queue, `dispose_item_function` shall be called with `dispose_item_function_context` and a pointer to the array entry value (T*). **]**
 
 ### TQUEUE_PUSH(T)
 ```c
-TQUEUE_PUSH_RESULT TQUEUE_PUSH(T)(TQUEUE(T) tqueue, const T item, void* push_cb_function_context)
+TQUEUE_PUSH_RESULT TQUEUE_PUSH(T)(TQUEUE(T) tqueue, T* item, void* push_cb_function_context)
 ```
 
 `TQUEUE_PUSH(T)` pushes an item in the queue.
```
anporumb, line 213: >. [](http://example.com/codeflow?start=44&length=1)

if this specifies "at the head" would be glorioous.

inc/clds/tqueue_ll.h
```diff
@@ -4,6 +4,16 @@
 #ifndef TQUEUE_LL_H
 #define TQUEUE_LL_H
 
+#ifdef __cplusplus
+#include <cstdbool>
```
anporumb, line 8: gotta ask where is this needed...

anporumb: As a side note, we should target one of the C++ ISO standards, right now it is sort of ... unspecified.

inc/clds/tqueue_ll.h
```diff
@@ -23,17 +33,368 @@ MU_DEFINE_ENUM(TQUEUE_PUSH_RESULT, TQUEUE_PUSH_RESULT_VALUES);
...
+                    } while (1);                                                                                                                                    \
+                                                                                                                                                                    \
+                    if (tqueue->push_cb_function == NULL)                                                                                                           \
+                    {                                                                                                                                               \
+                        /* Codes_SRS_TQUEUE_01_019: [ If no push_cb_function was specified in TQUEUE_CREATE(T), TQUEUE_PUSH(T) shall copy the value of item into the array entry value whose state was changed to PUSHING. ]*/ \
+                        (void)memcpy((void*)(T*)&tqueue->queue[index].value, (void*)item, sizeof(T));                                                               \
```
anporumb, line 268: >(T*) [](http://example.com/codeflow?start=44&length=4)

likely cast not needed.

anporumb, line 268: >(void*) [](http://example.com/codeflow?start=37&length=7)

why is this cast needed?

inc/clds/tqueue_ll.h
```diff
@@ -23,17 +33,368 @@ MU_DEFINE_ENUM(TQUEUE_PUSH_RESULT, TQUEUE_PUSH_RESULT_VALUES);
...
+                            /* Codes_SRS_TQUEUE_01_023: [ If the state of the array entry corresponding to the head is not NOT_USED, TQUEUE_PUSH(T) shall try changing the state again. ]*/ \
+                        }                                                                                                                                           \
+                        else                                                                                                                                        \
+                        {                                                                                                                                           \
+                            break;                                                                                                                                  \
+                        }                                                                                                                                           \
+                    } while (1);                                                                                                                                    \
```
anporumb, line 263: >while (1); [](http://example.com/codeflow?start=22&length=10)

I am wondering if line 261 did not execute... what would get the loop to exit

For line 261 to not execute, it would mean that the state of the item at head would not be "NOT_USED". So ... assume it is "USED" (for giggles).

If it was used, then the queue sort of is full and ... something doesn't work some other place.

For the item's state at "index" to be in a state different than "NOT_USED" it would mean somebody else has ownership over that index. But since this is wanna-be head (and we don't get out until the item gets unused), then... 

1) thread will loop forever here (eating CPU time, big bad wolf, not good)
2) queue is full and the retry should be done from line 228.

I think.

inc/clds/tqueue_ll.h
```diff
@@ -23,17 +33,368 @@ MU_DEFINE_ENUM(TQUEUE_PUSH_RESULT, TQUEUE_PUSH_RESULT_VALUES);
...
+
+/*TQUEUE_DEFINE_STRUCT_TYPE(T) introduces the base type that holds the queue typed as T*/
+#define TQUEUE_DEFINE_STRUCT_TYPE(T)                                                                            \
+typedef void (*TQUEUE_DEFINE_PUSH_CB_FUNCTION_TYPE_NAME(T))(void* context, T* push_dst, T* push_src);           \
+typedef void (*TQUEUE_DEFINE_POP_CB_FUNCTION_TYPE_NAME(T))(void* context, T* pop_dst, T* pop_src);              \
+typedef void (*TQUEUE_DEFINE_DISPOSE_ITEM_FUNCTION_TYPE_NAME(T))(void* context, T* item);                       \
+typedef bool (*TQUEUE_DEFINE_CONDITION_FUNCTION_TYPE_NAME(T))(void* context, T* item);                          \
+typedef struct TQUEUE_ENTRY_STRUCT_TYPE_NAME_TAG(T)                                                             \
+{                                                                                                               \
+    volatile_atomic int32_t state;                                                                              \
```
anporumb, line 75: >volatile_atomic int32_t state; [](http://example.com/codeflow?start=4&length=30)

suggest to have an anonymous union here over "int32_t and QUEUE_ENTRY_STATE" for debug purposes.

inc/clds/tqueue_ll.h
```diff
@@ -23,17 +33,368 @@ MU_DEFINE_ENUM(TQUEUE_PUSH_RESULT, TQUEUE_PUSH_RESULT_VALUES);
...
+                    {                                                                                                                                               \
+                        /* Codes_SRS_TQUEUE_01_024: [ If a push_cb_function was specified in TQUEUE_CREATE(T), TQUEUE_PUSH(T) shall call the push_cb_function with push_cb_function_context as context, a pointer to the array entry value whose state was changed to PUSHING as push_dst and item as push_src. ] */ \
+                        tqueue->push_cb_function(push_cb_function_context, (T*)&tqueue->queue[index].value, item);                                                  \
+                    }                                                                                                                                               \
+                    /* Codes_SRS_TQUEUE_01_020: [ TQUEUE_PUSH(T) shall set the state to USED by using interlocked_exchange. ]*/                                     \
+                    (void)interlocked_exchange((volatile_atomic int32_t*)&tqueue->queue[index].state, QUEUE_ENTRY_STATE_USED);                                      \
```
anporumb, line 276: >(volatile_atomic int32_t*) [](http://example.com/codeflow?start=47&length=26)

why is this cast needed?

inc/clds/tqueue_ll.h
```diff
@@ -23,17 +33,368 @@ MU_DEFINE_ENUM(TQUEUE_PUSH_RESULT, TQUEUE_PUSH_RESULT_VALUES);
...
+                /* LogInfo("Popping from %" PRId64 "", current_tail); */ \
+                /* Codes_SRS_TQUEUE_01_030: [ Using interlocked_compare_exchange, TQUEUE_PUSH(T) shall set the tail array entry state to POPPING (from USED). ]*/   \
+                uint32_t index = (uint32_t)current_tail % tqueue->queue_size;                                                                                       \
+                if (interlocked_compare_exchange((volatile_atomic int32_t*)&tqueue->queue[index].state, QUEUE_ENTRY_STATE_POPPING, QUEUE_ENTRY_STATE_USED) != QUEUE_ENTRY_STATE_USED) \
+                {                                                                                                                                                   \
+                    /* Codes_SRS_TQUEUE_01_036: [ If the state of the array entry corresponding to the tail is not USED, TQUEUE_POP(T) shall try again. ]*/         \
+                    continue;                                                                                                                                       \
```
anporumb, line 326: >continue [](http://example.com/codeflow?start=20&length=8)

So POP gets a redo (goes all the way to the begiining" while PUSH gets a do/while(1).

They are not so different... one of them is wrong.

inc/clds/tqueue_ll.h
```diff
@@ -23,17 +33,368 @@ MU_DEFINE_ENUM(TQUEUE_PUSH_RESULT, TQUEUE_PUSH_RESULT_VALUES);
...
+                    bool should_pop;                                                                                                                                \
+                    /* Codes_SRS_TQUEUE_01_039: [ If condition_function is not NULL: ]*/                                                                            \
+                    if (condition_function != NULL)                                                                                                                 \
+                    {                                                                                                                                               \
+                        /* Codes_SRS_TQUEUE_01_040: [ TQUEUE_POP(T) shall call condition_function with condition_function_context and a pointer to the array entry value whose state was changed to POPPING. ] */ \
+                        should_pop = condition_function(condition_function_context, (T*)&tqueue->queue[index].value);                                                   \
```
anporumb, line 335: >(T*) [](http://example.com/codeflow?start=84&length=4)

w2hy is this cast needed?

inc/clds/tqueue_ll.h
```diff
@@ -23,17 +33,368 @@ MU_DEFINE_ENUM(TQUEUE_PUSH_RESULT, TQUEUE_PUSH_RESULT_VALUES);
...
+                        should_pop = true;                                                                                                                          \
+                    }                                                                                                                                               \
+                    if (!should_pop)                                                                                                                                \
+                    {                                                                                                                                               \
+                        /* Codes_SRS_TQUEUE_01_041: [ If condition_function returns false, TQUEUE_POP(T) shall set the state to USED by using interlocked_exchange and return TQUEUE_POP_REJECTED. ]*/ \
+                        (void)interlocked_exchange((volatile_atomic int32_t*)&tqueue->queue[index].state, QUEUE_ENTRY_STATE_USED);                                  \
```
anporumb, line 345: >(volatile_atomic int32_t*) [](http://example.com/codeflow?start=51&length=26)

why is cast needed?

inc/clds/tqueue_ll.h
```diff
@@ -23,17 +33,368 @@ MU_DEFINE_ENUM(TQUEUE_PUSH_RESULT, TQUEUE_PUSH_RESULT_VALUES);
...
+                        else                                                                                                                                        \
+                        {                                                                                                                                           \
+                            if (tqueue->pop_cb_function == NULL)                                                                                                    \
+                            {                                                                                                                                       \
+                                /* Codes_SRS_TQUEUE_01_032: [ If a pop_cb_function was not specified in TQUEUE_CREATE(T): ]*/                                       \
+                                /* Codes_SRS_TQUEUE_01_033: [ TQUEUE_POP(T) shall copy array entry value whose state was changed to POPPING to item. ]*/            \
+                                (void)memcpy((void*)item, (void*)(T*)&tqueue->queue[index].value, sizeof(T));                                                                    \
```
anporumb, line 363: >(void*) [](http://example.com/codeflow?start=45&length=7)

why is this cast needed?

anporumb, line 363: >(T*) [](http://example.com/codeflow?start=65&length=4)

why is this cast needed?

anporumb, line 363: >sizeof(T) [](http://example.com/codeflow?start=98&length=9)

cough.

Now sizeof(T) is all nice, but what does sizeof(T) return for a struct that has a flexible array member? 

Here (and everywhere where sizeof(T)) is used I believe we need some oooooooother sort of CB that computes the size in bytes to be copied.

Maybe a next generation CR...

inc/clds/tqueue_ll.h
```diff
@@ -23,17 +33,368 @@ MU_DEFINE_ENUM(TQUEUE_PUSH_RESULT, TQUEUE_PUSH_RESULT_VALUES);
...
+                        /* Codes_SRS_TQUEUE_01_031: [ TQUEUE_POP(T) shall replace the tail value with the tail value obtained earlier + 1 by using interlocked_exchange_64. ]*/ \
+                        if (interlocked_compare_exchange_64((volatile_atomic int64_t*)&tqueue->tail, current_tail + 1, current_tail) != current_tail)               \
+                        {                                                                                                                                           \
+                            /* Codes_SRS_TQUEUE_01_044: [ If incrementing the tail by using interlocked_compare_exchange_64 does not succeed, TQUEUE_POP(T) shall revert the state of the array entry to USED and retry. ]*/ \
+                            (void)interlocked_exchange((volatile_atomic int32_t*)&tqueue->queue[index].state, QUEUE_ENTRY_STATE_USED);                              \
+                            continue;                                                                                                                               \
```
anporumb, line 355: >continue [](http://example.com/codeflow?start=28&length=8)

:(

So this is not friendly to the user. Because the dialogue between the user and the queue goes like (assuming user has a pop_if function)

WE: "hey user, here's a juicy item from the queue, do you want it?"
USER: "YES I DO".
[... then line 351 gets executed... so everything restarts]
WE: "hey user, here's a juicy item from the queue, do you want it?"
USER: "but... what happened to the previous item, why can't I have it? since you've asked me if I want it".

So I think once the user returns "true" from the "pop_if" condition we HAVE TO deliver the same item to the user.

inc/clds/tqueue_ll.h
anporumb, line 363: >sizeof(T) [](http://example.com/codeflow?start=98&length=9)

or documentation should make a note that flexible arrays are only useable with pop_copy_functions.

**Or not useable at all** for T and leave it at that. Not sure how it can even be detected at compile time.

inc/clds/tqueue_ll.h
```diff
@@ -23,17 +33,368 @@ MU_DEFINE_ENUM(TQUEUE_PUSH_RESULT, TQUEUE_PUSH_RESULT_VALUES);
...
+                                /* Codes_SRS_TQUEUE_01_037: [ If a pop_cb_function was specified in TQUEUE_CREATE(T): ]*/                                           \
+                                /* Codes_SRS_TQUEUE_01_038: [ TQUEUE_POP(T) shall call pop_cb_function with pop_cb_function_context as context, the array entry value whose state was changed to POPPING to item as pop_src and item as pop_dst. ]*/ \
+                                tqueue->pop_cb_function(pop_cb_function_context, item, (T*)&tqueue->queue[index].value);                                            \
+                            }                                                                                                                                       \
+                            /* Codes_SRS_TQUEUE_01_034: [ TQUEUE_POP(T) shall set the state to NOT_USED by using interlocked_exchange, succeed and return TQUEUE_POP_OK. ]*/ \
+                            (void)interlocked_exchange((volatile_atomic int32_t*)&tqueue->queue[index].state, QUEUE_ENTRY_STATE_NOT_USED);                          \
```
anporumb, line 372: >(volatile_atomic int32_t*) [](http://example.com/codeflow?start=55&length=26)

why is cast needed?

inc/clds/tqueue_ll.h
```diff
@@ -23,17 +33,368 @@ MU_DEFINE_ENUM(TQUEUE_PUSH_RESULT, TQUEUE_PUSH_RESULT_VALUES);
...
+                    }                                                                                                                                               \
+                    break;                                                                                                                                          \
+                }                                                                                                                                                   \
+            }                                                                                                                                                       \
+        } while (1);                                                                                                                                                \
+    }                                                                                                                                                               \
+    return result;                                                                                                                                                  \
```
anporumb, line 381: I think compilers go smarter about returning non-initialized values.

src/tqueue.c
```diff
@@ -11,3 +11,6 @@
 #include "c_pal/gballoc_hl_redirect.h"
```
anporumb, line 11: remove, not used.,

tests/tqueue_int/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+set(${theseTestsName}_c_files
+tqueue_foo.c
+)
+
+set(${theseTestsName}_h_files
+../../inc/clds/tqueue.h
+tqueue_foo.h
+)
+
+build_test_artifacts(${theseTestsName} "tests/clds" ADDITIONAL_LIBS clds c_pal c_pal_reals c_util c_util_reals)
```
anporumb, line 18: >c_pal_reals [](http://example.com/codeflow?start=79&length=11)

why are reals needed in int tests?

tests/tqueue_int/tqueue_int.c
```diff
@@ -0,0 +1,576 @@
...
+        switch (action_type)
+        {
+        default:
+            ASSERT_FAIL("Unexpected action type %" PRI_MU_ENUM "", MU_ENUM_VALUE(TQUEUE_ACTION_TYPE, action_type));
+            break;
+
+        case TQUEUE_ACTION_TYPE_PUSH:
+        {
+            int64_t next_push_number = interlocked_increment_64(&test_context->next_push_number);
+            FOO item = { .x = next_push_number };
```
anporumb, line 183: >.x [](http://example.com/codeflow?start=25&length=2)

this designated initializer is nice, can we use it everywhere?>

tests/tqueue_int/tqueue_int.c
```diff
@@ -0,0 +1,576 @@
...
+            {
+                (void)interlocked_increment_64(&test_context->succesful_push_count);
+            }
+            break;
+        }
+        case TQUEUE_ACTION_TYPE_POP:
+        {
+            FOO item = { .x = -1 };
+            TQUEUE_POP_RESULT pop_result = TQUEUE_POP(FOO)(test_context->queue, &item, NULL, NULL, NULL);
+            ASSERT_IS_TRUE((pop_result == TQUEUE_POP_OK) || (pop_result == TQUEUE_POP_QUEUE_EMPTY), "TQUEUE_POP(FOO) failed with %" PRI_MU_ENUM "", MU_ENUM_VALUE(TQUEUE_POP_RESULT, pop_result));
```
anporumb, line 196: >ASSERT_IS_TRUE [](http://example.com/codeflow?start=12&length=14)

we can do more asserts here.

First assert is that ".x" is not -1 (has been changed).

Another assert is that "this POP is greater than last POP".

tests/tqueue_int/tqueue_int.c
```diff
@@ -0,0 +1,576 @@
...
+
+static volatile_atomic int32_t terminate_test;
+
+static int tqueue_chaos_thread_func(void* arg)
+{
+    TQUEUE_CHAOS_TEST_CONTEXT* test_context = arg;
+
+    while (interlocked_add(&terminate_test, 0) == 0)
+    {
+        TQUEUE_ACTION_TYPE action_type = (TQUEUE_ACTION_TYPE)((uint32_t)rand() * ((MU_COUNT_ARG(TQUEUE_ACTION_TYPE_VALUES))) / ((uint32_t)RAND_MAX + 1));
```
anporumb, line 172: >action_type [](http://example.com/codeflow?start=27&length=11)

I believe we should have 3 flavors of chaos here:

1) is the one we have, perfectly balanced like all things should be

2) is one which is skewed towards pushers (have 66% of the operations being push)

3) is one which is skewed towards popers (have 66% of the operations being pop).

anporumb: Would also be nice to observe CPU time during chaos.

tests/tqueue_int/tqueue_int.c
```diff
@@ -0,0 +1,576 @@
...
+} TQUEUE_CHAOS_TEST_CONTEXT;
+
+#define TQUEUE_ACTION_TYPE_VALUES \
+    TQUEUE_ACTION_TYPE_PUSH, \
+    TQUEUE_ACTION_TYPE_POP \
+
+MU_DEFINE_ENUM_WITHOUT_INVALID(TQUEUE_ACTION_TYPE, TQUEUE_ACTION_TYPE_VALUES);
+MU_DEFINE_ENUM_STRINGS_WITHOUT_INVALID(TQUEUE_ACTION_TYPE, TQUEUE_ACTION_TYPE_VALUES);
+
+static volatile_atomic int32_t terminate_test;
```
anporumb, line 164: >terminate_test [](http://example.com/codeflow?start=31&length=14)

could be moved in `TQUEUE_CHAOS_TEST_CONTEXT`

tests/tqueue_int/tqueue_int.c
```diff
@@ -0,0 +1,576 @@
...
+        ASSERT_IS_TRUE(current_succesful_push_count > last_succesful_push_count);
+        ASSERT_IS_TRUE(current_succesful_pop_count > last_succesful_pop_count);
+
+        current_time = timer_global_get_elapsed_ms();
+
+        LogInfo("%.02f seconds elapsed, successful push count=%" PRId64 ", successful pop count=%" PRId64 "",
+            (current_time - start_time) / 1000, current_succesful_push_count, current_succesful_pop_count);
+    } while (current_time - start_time < CHAOS_TEST_RUNTIME);
+
+    ThreadAPI_Sleep(CHAOS_TEST_RUNTIME);
```
anporumb, line 272: >CHAOS_TEST_RUNTIME [](http://example.com/codeflow?start=20&length=18)

why does it need to sleep here?

tests/tqueue_int/tqueue_int.c
```diff
@@ -0,0 +1,576 @@
...
+        // yield
+        ThreadAPI_Sleep(0);
+#endif
+    }
+
+    return 0;
+}
+
+// This test has one pusher and one popper and validates the fact that order is preserved in this case
+TEST_FUNCTION(TQUEUE_test_with_1_pusher_and_1_popper)
```
anporumb, line 349: >TQUEUE_test_with_1_pusher_and_1_popper [](http://example.com/codeflow?start=14&length=38)

torture is when "1_pusher_and_1_popper_and_1_queue_size"

tests/tqueue_int/tqueue_int.c
```diff
@@ -0,0 +1,576 @@
...
+    ASSERT_ARE_EQUAL(THREADAPI_RESULT, THREADAPI_OK, ThreadAPI_Create(&pusher_thread, pusher_thread_func, &test_context));
+    ASSERT_ARE_EQUAL(THREADAPI_RESULT, THREADAPI_OK, ThreadAPI_Create(&popper_thread, popper_thread_func, &test_context));
+
+    double start_time = timer_global_get_elapsed_ms();
+    double current_time = start_time;
+    do
+    {
+        int64_t last_expected_pop_number = interlocked_add_64(&test_context.next_expected_pop_number, 0);
+
+        ThreadAPI_Sleep(TEST_CHECK_PERIOD);
```
anporumb, line 374: >ThreadAPI_Sleep(TEST_CHECK_PERIOD); [](http://example.com/codeflow?start=8&length=35)

why does it need to sleep here?

tests/tqueue_int/tqueue_int.c
```diff
@@ -0,0 +1,576 @@
...
+    (void)interlocked_exchange(&terminate_test, 1);
+
+    {
+        int dont_care;
+        ASSERT_ARE_EQUAL(THREADAPI_RESULT, THREADAPI_OK, ThreadAPI_Join(pusher_thread, &dont_care));
+    }
+
+    {
+        int dont_care;
+        ASSERT_ARE_EQUAL(THREADAPI_RESULT, THREADAPI_OK, ThreadAPI_Join(popper_thread, &dont_care));
```
anporumb, line 395: suspicion of copy paste

tests/tqueue_int/tqueue_int.c
```diff
@@ -0,0 +1,576 @@
...
+        int64_t current_succesful_push_count = interlocked_add_64(&test_context.succesful_push_count, 0);
+        int64_t current_succesful_pop_count = interlocked_add_64(&test_context.succesful_pop_count, 0);
+
+        // make sure we had at least one succesful push and one pop (not stuck)
+        ASSERT_IS_TRUE(current_succesful_push_count > last_succesful_push_count);
+        ASSERT_IS_TRUE(current_succesful_pop_count > last_succesful_pop_count);
+
+        current_time = timer_global_get_elapsed_ms();
+
+        LogInfo("%.02f seconds elapsed, successful push count=%" PRId64 ", successful pop count=%" PRId64 "",
```
anporumb, line 557: >PRId64 [](http://example.com/codeflow?start=99&length=6)

If these numbers get very big... suggest using PRI_KB :) [shamelessly bumping some of my own work]

tests/tqueue_int/tqueue_int.c
```diff
@@ -0,0 +1,576 @@
...
+    {
+        int dont_care;
+        ASSERT_ARE_EQUAL(THREADAPI_RESULT, THREADAPI_OK, ThreadAPI_Join(thread_handles[i], &dont_care));
+    }
+
+    // clean
+    TQUEUE_ASSIGN(THANDLE(TEST_THANDLE))(&test_context.queue, NULL);
+}
+
+END_TEST_SUITE(TEST_SUITE_NAME_FROM_CMAKE)
```
anporumb, line 576: >TEST_SUITE_NAME_FROM_CMAKE [](http://example.com/codeflow?start=15&length=26)

So ending commentary here,

I believe we should have chaos test which are unbalanced, and chaos tests that use just 1 queue slot (with many competing threads).

I see already THANDLE tests, cheers!

tests/tqueue_ut/tqueue_ut.c
```diff
@@ -0,0 +1,1081 @@
...
+
+BEGIN_TEST_SUITE(TEST_SUITE_NAME_FROM_CMAKE)
+
+TEST_SUITE_INITIALIZE(suite_init)
+{
+    ASSERT_ARE_EQUAL(int, 0, real_gballoc_hl_init(NULL, NULL));
+
+    ASSERT_ARE_EQUAL(int, 0, umock_c_init(on_umock_c_error));
+    ASSERT_ARE_EQUAL(int, 0, umocktypes_bool_register_types(), "umocktypes_bool_register_types failed");
+    ASSERT_ARE_EQUAL(int, 0, umocktypes_charptr_register_types(), "umocktypes_charptr_register_types failed");
```
anporumb, line 79: >umocktypes_charptr_register_types [](http://example.com/codeflow?start=29&length=33)

liiiiikely not needed.

tests/tqueue_ut/tqueue_ut.c
```diff
@@ -0,0 +1,1081 @@
...
+    STRICT_EXPECTED_CALL(interlocked_exchange(IGNORED_ARG, 1));
+    STRICT_EXPECTED_CALL(interlocked_exchange_64(IGNORED_ARG, 0));
+    STRICT_EXPECTED_CALL(interlocked_exchange_64(IGNORED_ARG, 0));
+    for (uint32_t i = 0; i < 1024; i++)
+    {
+        STRICT_EXPECTED_CALL(interlocked_exchange(IGNORED_ARG, QUEUE_ENTRY_STATE_NOT_USED));
+    }
+
+    // act
+    TQUEUE(int32_t) queue = TQUEUE_CREATE(int32_t)(1024, NULL, NULL, NULL, (void*)0x4242);
```
anporumb, line 231: >(void*)0x4242 [](http://example.com/codeflow?start=75&length=13)

suggest "NULL" cause it is not used anyway.

devdoc/tqueue_requirements.md
anporumb, line 112: >POP_CB_FUNCTION [](http://example.com/codeflow?start=19&length=15)

I still don't know why we need a POP_CB and a PUSH_CB which are different.

What's a use-case where POP_CB and PUSH_CB need to be different instead of "COPY_ITEM"? Once it is copied "from" the Q to the user, the other time it is copied from the USER to the Q. 

but since these 2 callbacks are user-implemented... and user is not aware of any implementation details of the Q, then why are they different?

tests/tqueue_ut/tqueue_ut.c
```diff
@@ -0,0 +1,1081 @@
...
+    // assert
+    ASSERT_ARE_EQUAL(TQUEUE_POP_RESULT, TQUEUE_POP_OK, result);
+    ASSERT_ARE_EQUAL(int32_t, 42, item);
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+
+    // clean
+    TQUEUE_ASSIGN(int32_t)(&queue, NULL);
+}
+
+END_TEST_SUITE(TEST_SUITE_NAME_FROM_CMAKE)
```
anporumb, line 1081: >TEST_SUITE_NAME_FROM_CMAKE [](http://example.com/codeflow?start=15&length=26)

I am curious.

I am also very busy.

I will sign off and if there's something that I could look at, ping me vigorously next iteration.

anporumb (approved):
:shipit:

devdoc/tqueue_requirements.md
dcristoloveanu: Fixed.

devdoc/tqueue_requirements.md
dcristoloveanu: Hehe, I had it const, but const const did not play well with THANDLE being the type for the TQUEUE.
I'll file a task to tinker with the constness at some point, as I do want to be able to have TQUEUE of THANDLE more than this... Ideally we want both obviously, as const is cool, but oh well, only 2 hands and a rake.

dcristoloveanu: Standby, I can make them one and call copy+dispose item when popping.
That way we have one less callback.

dcristoloveanu: No longer we have push_cb and pop_cb. Now we have copy item and dispose_item functions.

devdoc/tqueue_requirements.md
dcristoloveanu: I created a task to attempt to fix this somehow and still work with THANDLE being stored in a TQUEUE.

devdoc/tqueue_requirements.md
dcristoloveanu: This was supposed to be the condition function (which returns a bool). Fixed now to have the signature of the condition function here.

devdoc/tqueue_requirements.md
dcristoloveanu: Yes, it is, now this has been cleaned up.

devdoc/tqueue_requirements.md
dcristoloveanu: Now we have copy_item and dispose_item functions, not longer with _cb

devdoc/tqueue_requirements.md
dcristoloveanu: Added gloriousness

inc/clds/tqueue_ll.h
dcristoloveanu: Not needed, removed. Was added due to inertia of adding the C header.

inc/clds/tqueue_ll.h
dcristoloveanu: It builds without so happy to not have it.

dcristoloveanu: Because if T is a THANDLE then const gets in the way. I added a task to see if I can make things better in that regard.

inc/clds/tqueue_ll.h
dcristoloveanu: Added

inc/clds/tqueue_ll.h
dcristoloveanu: Needed because const from the TQUEUE gets in the way.

inc/clds/tqueue_ll.h
```diff
@@ -23,17 +33,368 @@ MU_DEFINE_ENUM(TQUEUE_PUSH_RESULT, TQUEUE_PUSH_RESULT_VALUES);
...
+        do                                                                                                                                                          \
+        {                                                                                                                                                           \
+            /* Codes_SRS_TQUEUE_01_028: [ TQUEUE_POP(T) shall obtain the current head queue by calling interlocked_add_64. ]*/                                      \
+            int64_t current_head = interlocked_add_64((volatile_atomic int64_t*)&tqueue->head, 0);                                                                  \
+            /* Codes_SRS_TQUEUE_01_029: [ TQUEUE_POP(T) shall obtain the current tail queue by calling interlocked_add_64. ]*/                                      \
+            int64_t current_tail = interlocked_add_64((volatile_atomic int64_t*)&tqueue->tail, 0);                                                                  \
+            if (current_tail >= current_head)                                                                                                                       \
```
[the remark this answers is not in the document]
dcristoloveanu: True, just put it there defensively.

inc/clds/tqueue_ll.h
```diff
@@ -23,17 +33,368 @@ MU_DEFINE_ENUM(TQUEUE_PUSH_RESULT, TQUEUE_PUSH_RESULT_VALUES);
...
+            {                                                                                                                                                       \
+                /* Codes_SRS_TQUEUE_01_035: [ If the queue is empty (current tail == current head), TQUEUE_POP(T) shall return TQUEUE_POP_QUEUE_EMPTY. ]*/          \
```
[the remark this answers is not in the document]
dcristoloveanu: Updated in the spec.

inc/clds/tqueue_ll.h
dcristoloveanu: Because of const of the tqueue

inc/clds/tqueue_ll.h
dcristoloveanu: If T is const, like in the case of a THANDLE.

inc/clds/tqueue_ll.h
dcristoloveanu: Because T can be a THANDLE.

inc/clds/tqueue_ll.h
dcristoloveanu: All of them casts are because of const because of the case when T is a THANDLE

dcristoloveanu: Yeah, I think this would be a flavor of the TQUEUE. Has to come later as a different task ...

dcristoloveanu: [https://msazure.visualstudio.com/One/_sprints/taskboard/Azure%20Messaging%20Store/One/Custom/AzureMessaging/Germanium/GeM1?workitem=25661413](https://msazure.visualstudio.com/One/_sprints/taskboard/Azure%20Messaging%20Store/One/Custom/AzureMessaging/Germanium/GeM1?workitem=25661413)

dcristoloveanu: It can probably be done with another function, I have added a task for it.

inc/clds/tqueue_ll.h
dcristoloveanu: Because T can be a THANDLE.

src/tqueue.c
dcristoloveanu: Removed

tests/CMakeLists.txt
```diff
@@ -37,6 +38,7 @@ endif()
     build_test_folder(clds_sorted_list_int)
     build_test_folder(mpsc_lock_free_queue_int)
     build_test_folder(tcall_dispatcher_int)
+    build_test_folder(tqueue_int)
```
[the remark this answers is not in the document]
dcristoloveanu: Adding a task for it.
[https://msazure.visualstudio.com/One/_sprints/taskboard/Azure%20Messaging%20Store/One/Custom/AzureMessaging/Germanium/GeM1?workitem=25661442](https://msazure.visualstudio.com/One/_sprints/taskboard/Azure%20Messaging%20Store/One/Custom/AzureMessaging/Germanium/GeM1?workitem=25661442)

tests/tqueue_int/CMakeLists.txt
dcristoloveanu: Not needed, removed

tests/tqueue_int/tqueue_int.c
dcristoloveanu: Moved to test_context.

tests/tqueue_ut/tqueue_ut.c
dcristoloveanu: Removed

tests/tqueue_int/tqueue_int.c
dcristoloveanu: This is waiting for the popper thread, does not look like copy/paste issue to me.

tests/tqueue_ut/tqueue_ut.c
dcristoloveanu: NULL-ified.

tests/tqueue_int/tqueue_int.c
dcristoloveanu: No particular reason, just wanted to check every some time that there was some progress. Just a simple basic yield.

tests/tqueue_int/tqueue_int.c
dcristoloveanu: I actually like them to be accurate and not truncated with KB/MB, etc. I would not be able to tell the difference between 1MB and 1MB if they are different by let's say 10.

dcristoloveanu: Not that I don't like your bumped work , but in a place where it is warranted :-).

tests/tqueue_int/tqueue_int.c
dcristoloveanu: Actually it does not lol. Was leftover from the 1st iteration of the test.

tests/tqueue_int/tqueue_int.c
dcristoloveanu: Added

tests/tqueue_int/tqueue_int.c
dcristoloveanu: Added check for not -1.
The other check (this POP is greater than last POP) is not necessarily true because this is multiple producers.

tests/tqueue_int/tqueue_int.c
dcristoloveanu: Used it now as much as I can in this file at least.

inc/clds/tqueue_ll.h
dcristoloveanu: I am not sure what this comments refers too. I looked at all code paths and all of them set result to something. ..

tests/tqueue_int/tqueue_int.c
dcristoloveanu: 1 queue slot has been added.

dcristoloveanu: Added also flavors that are imbalanced (1 pusher with n poppers and the other way around).

tests/tqueue_int/tqueue_int.c
dcristoloveanu: Added the flavors of chaos.
As for CPU that will have to come as a separate thing. I did discuss at some point with Jelani about that - we needed it for another int test, but doing that now is not helping ...

inc/clds/tqueue_ll.h
dcristoloveanu: Changed the order of the changes for state and head, so now both ends do:
- change state
- change tail/head

I spent several hours thinking and working through a better algo. There is one, so stay on the edge of your seat, but frankly I will do that separately as I'd like to do that not under pressure.

inc/clds/tqueue_ll.h
dcristoloveanu: They have the same order now

inc/clds/tqueue_ll.h
dcristoloveanu: I know! There is a better algo (read moving the tail after delivery combined with some nice stamping of the items with the lsns for invariants). That would fix this I believe, but I need the time for it and now we have to move with the fix.

tests/tqueue_ut/tqueue_ut.c
dcristoloveanu: I believe the biggest concern should be addressed (the spin in push), the rest have to come with other tasks.

avranju (approved):

PR #205, opened 2023-11-02
A fix (not the best) for the perf degradation of HP

dcristoloveanu:


src/clds_hazard_pointers_thread_helper.c
```diff
@@ -9,17 +9,73 @@
...
 
 typedef struct CLDS_HAZARD_POINTERS_THREAD_HELPER_TAG
 {
     CLDS_HAZARD_POINTERS_HANDLE hazard_pointers;
     DWORD tls_slot;
+    TCALL_DISPATCHER_TARGET_HANDLE(THREAD_NOTIFICATION_CALL) thread_notification_target_handle;
+    TCALL_DISPATCHER(THREAD_NOTIFICATION_CALL) call_dispatcher;
 } CLDS_HAZARD_POINTERS_THREAD_HELPER;
 
+MU_DEFINE_ENUM_STRINGS(THREAD_NOTIFICATIONS_LACKEY_DLL_REASON, THREAD_NOTIFICATIONS_LACKEY_DLL_REASON_VALUES);
```
mattdurak, line 27: >THREAD_NOTIFICATIONS_LACKEY_DLL_REASON [](http://example.com/codeflow?start=23&length=38)

nit: I imagine this belongs in some other .c file matching the .h file that the enum came from

src/clds_hazard_pointers_thread_helper.c
```diff
@@ -42,20 +98,45 @@ IMPLEMENT_MOCKABLE_FUNCTION(, CLDS_HAZARD_POINTERS_THREAD_HELPER_HANDLE, clds_ha
...
+                /* Codes_SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_01_002: [ clds_hazard_pointers_thread_helper_create shall register clds_hazard_pointers_thread_helper_thread_notification call target with the TCALL_DISPATCHER(THREAD_NOTIFICATION_CALL). ]*/
+                result->thread_notification_target_handle = TCALL_DISPATCHER_REGISTER_TARGET(THREAD_NOTIFICATION_CALL)(result->call_dispatcher, clds_hazard_pointers_thread_helper_thread_notification, result);
+                if (result->thread_notification_target_handle == NULL)
+                {
+                    LogError("thread_notifications_register_notification failed");
+                }
+                else
+                {
+                    /*Codes_SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_42_003: [ clds_hazard_pointers_thread_helper_create shall allocate the thread local storage slot for the hazard pointers by calling TlsAlloc. ]*/
+                    result->tls_slot = TlsAlloc();
```
mattdurak, line 121: >TlsAlloc [](http://example.com/codeflow?start=39&length=8)

I think this should happen before registering, otherwise the callback could come from some random thread exiting and try to do TlsGetValue on some garbage (which may or may not be a valid slot).

tests/clds_hazard_pointers_thread_helper_perf/clds_hazard_pointers_thread_helper_perf.c
```diff
@@ -228,7 +232,6 @@ static void test_skipped_seq_chaos(void* context, int64_t skipped_sequence_no)
 }
 
 // Test does not pass until we fix the scalability issue with no. of threads used
```
mattdurak, line 234: likely the comment goes away

tests/clds_hazard_pointers_thread_helper_ut/clds_hazard_pointers_thread_helper_ut.c
```diff
@@ -343,4 +413,109 @@ TEST_FUNCTION(clds_hazard_pointers_thread_helper_get_thread_fails_when_TlsSetVal
...
+
+    // cleanup
+    clds_hazard_pointers_thread_helper_destroy(helper);
+}
+
+/* Tests_SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_01_007: [ If reason is THREAD_NOTIFICATIONS_LACKEY_DLL_REASON_THREAD_DETACH, clds_hazard_pointers_thread_helper_thread_notification shall call TlsGetValue obtain the thread local value for the slot created in the clds_hazard_pointers_thread_create. ]*/
+/* Tests_SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_01_009: [ If the thread local stored value is not NULL: ]*/
+/* Tests_SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_01_010: [ clds_hazard_pointers_thread_helper_thread_notification shall set the value in the slot to NULL by calling TlsSetValue. ]*/
+/* Tests_SRS_CLDS_HAZARD_POINTERS_THREAD_HELPER_01_011: [ clds_hazard_pointers_thread_helper_thread_notification shall call clds_hazard_pointers_unregister_thread with the argument being the obtained thread local value. ]*/
+TEST_FUNCTION(clds_hazard_pointers_thread_helper_thread_notification_with_DETACH_returns)
```
mattdurak, line 460: >returns) [](http://example.com/codeflow?start=81&length=8)

unregisters_thread?

src/clds_hazard_pointers.c
```diff
@@ -43,8 +58,18 @@ typedef struct CLDS_HAZARD_POINTERS_THREAD_TAG
 
 typedef struct CLDS_HAZARD_POINTERS_TAG
 {
+    WORKER_THREAD_HANDLE hp_thread_cleanup_worker;
     size_t reclaim_threshold;
     CLDS_HAZARD_POINTERS_THREAD* volatile_atomic head;
+    // This epoch exists in order to make sure that no HP thread information that is still being accessed
+    // is actually freed
+    // HP Thread data is only freed when it was placed in theinactive queue and its epoch is lower than the current epoch
```
mattdurak, line 66: >theinactive [](http://example.com/codeflow?start=58&length=11)

nit: missing space

src/clds_hazard_pointers.c
```diff
@@ -177,24 +275,119 @@ static void internal_reclaim(CLDS_HAZARD_POINTERS_THREAD_HANDLE clds_hazard_poin
...
+        if (interlocked_add(&current_thread->active, 0) == 1)
+        {
+            current_thread_address = (void* volatile_atomic*)&current_thread->next;
+            current_thread = next_thread;
+        }
+        else
+        {
+            // remove from the list, no longer active
+            if (interlocked_compare_exchange_pointer(current_thread_address, next_thread, current_thread) == current_thread)
+            {
```
mattdurak, line 317: Is this safe if next_thread is NULL?

E.g.

List is:
A -> B -> C -> NULL

C is inactive

We remove it and have:
A -> B -> NULL

BUT, what if another thread is added to the list between lines 307 and 316?

So we have "next=NULL"
Then the list actually becomes:
A -> B -> C -> D -> NULL

And line 316 loses D

So I think removing the last item in the list might not work

mattdurak: nevermind, I wasn't thinking about how the insert works, my example above should really be:

D -> A -> B -> C -> NULL

src/clds_hazard_pointers.c
```diff
@@ -72,9 +97,75 @@ static int hp_key_compare(void* key1, void* key2)
...
+        hazard_ptr = next_hazard_ptr;
+    }
+
+    free(clds_hazard_pointers_thread);
+}
+
+static bool is_epoch_lower_or_equal(void* context, CLDS_HP_INACTIVE_THREAD* inactive_thread)
+{
+    int64_t* epoch = context;
+    return inactive_thread->current_epoch_value <= *epoch;
```
mattdurak, line 145: Is "<=" correct, or should it be "<"?
Is it a problem if we free threads from the current epoch? Shouldn't we only free from "older" epochs?

mattdurak (approved):
:shipit:

src/clds_hazard_pointers_thread_helper.c
dcristoloveanu: Unfortunately that is in a dll, so I could not bring it easily in.

src/clds_hazard_pointers_thread_helper.c
dcristoloveanu: Good catch, fixing it

dcristoloveanu: Done.

tests/clds_hazard_pointers_thread_helper_perf/clds_hazard_pointers_thread_helper_perf.c
dcristoloveanu: Removed

tests/clds_hazard_pointers_thread_helper_ut/clds_hazard_pointers_thread_helper_ut.c
dcristoloveanu: Fixed.

src/clds_hazard_pointers.c
dcristoloveanu: Fixed

src/clds_hazard_pointers.c
dcristoloveanu: It should be correct, but I'll update to do < and change what we call it with to reduce confusion.

jebrando (approved):
:shipit:

PR #206, opened 2023-11-02
Add inactive_clds_hp_thread_queue

dcristoloveanu:
Add a simple queue (inactive_clds_hp_thread_queue) to be used in tracking inactive hp threads that are to be cleaned up.


avranju (approved):
LGTM.

PR #213, opened 2023-11-10
Fix reals for inactive_hp_thread_queue

dcristoloveanu:


BertKleewein (dismissed):
:shipit:

tests/reals/real_inactive_hp_thread_queue.c
```diff
@@ -6,6 +6,7 @@
 #include "c_pal/gballoc_hl_redirect.h"
```
anporumb, line 7: why is this needed here?

anporumb (dismissed):
:shipit:

tests/reals/real_inactive_hp_thread_queue.c
dcristoloveanu: Likely not needed, I am removing it and let's see where it craps out.

anporumb (approved):
:shipit:

PR #297, opened 2024-11-05
Reduce to half the time spent in hash table snapshot by using the item count maintained by the hash table to allocate the results array

dcristoloveanu:
Reduce to half the time spent in hash table snapshot by using the item count maintained by the hash table to allocate the results array


mattdurak (approved):
:shipit:

PR #311, opened 2024-11-28
Snapshot cancellation token

dcristoloveanu:
- Add a cancellation token in the snapshot API
- Add UTs
- Add int tests
- Add chaos with snapshot int test


jebrando (approved):
:shipit:

PR #312, opened 2024-12-04
Fix reals include for clds_hash_table

dcristoloveanu:
Fix reals include for clds_hash_table

### DIFF
--- a/tests/reals/real_clds_hash_table_renames.h
+++ b/tests/reals/real_clds_hash_table_renames.h
@@ -1,8 +1,6 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-#include "real_cancellation_token_renames.h"
-
 #define clds_hash_table_create real_clds_hash_table_create
 #define clds_hash_table_destroy real_clds_hash_table_destroy
 #define clds_hash_table_insert real_clds_hash_table_insert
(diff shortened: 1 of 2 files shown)


tests/reals/real_clds_hash_table_renames.h
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+#include "real_cancellation_token_renames.h"
```
anporumb, line 4: I will take issue with this... "these are not the droids you're looking for".

The purpose of the "_renames.h" file is to provide a list of all the symbols that need to be renames from *the current file* and rename them by means of pound defines.

These "XXX_renames.h" should follow the `#include "other_api.h"`, #include "api,.h" which are used in production C files.

Here's statistics:

EBS (and deps) has 60 such "real_MODULE_renames.h" files. Out of 60 files, only 5 have a #include in them (probably wrong all of them).

So I think these are not the droids you are looking for ./jedi hand wave.

Probably the "real_cancellation_token_renames.h" needs (or not) to be included in some other file, where currently "real_clds_hash_table_renames.h" included.

The reason why we should not include other renames is because they can sneakingly introduce renames when you might not expect those to appear. Let's keep "the module should only rename the symbols it own" rule.

anporumb:
> #define clds_hash_table_snapshot real_clds_hash_table_snapshot

file does not end with an empty line.

---
Refers to: tests/reals/real_clds_hash_table_renames.h:17 in cb6eac6. [](commit_id = cb6eac62f0df0ef1579b85186ac91b5117564e3d, deletion_comment = False)

anporumb (approved):
:shipit:

tests/reals/real_clds_hash_table_renames.h
dcristoloveanu: Moved the _renames include in the .c file where it belongs

mattdurak (approved):
:shipit:

PR #321, opened 2025-02-12
Fix for hitting limit of inactive threads by pruning the list more.

mattdurak:
Problem:
- When many threads are started and nothing is removed from the hash table, the inactive thread list can grow to the limit of 16k. When we hit this limit, we terminate the process.

Solution:
- When we are adding inactive threads, we also attempt to increment the hazard pointers epoch and cleanup old inactive threads.

Also add int test to catch the bug by spawning 32k threads that will become inactive without ever deleting an item from the hash table.


dcristoloveanu (approved):
:shipit:

PR #336, opened 2025-05-13
Fix doubling of memory for hash tables

dcristoloveanu:
- Add a test exposing the issue
- Fix the counting of the items for each level of the hash table


mattdurak (approved):
:shipit:

PR #408, opened 2026-04-03
[autogenerated] update deps: c-build-tools, c-pal

parth21999:
### c-build-tools
- `1a00465` Fix: guard .gitmodules access for repos without submodules (#376)
- `473d3d1` [MrBot] Remove validation_msbuild_target to enable automatic submodule propagation (#375)
- `ada7a79` Fix orphaned else in update-c-build-tools-yaml-refs (#367)
- `b92332c` [MrBot] Extract repo validation and traceability to run once instead of per-flavor (#369)

### c-pal
- `a88947a` [MrBot] Remove THANDLE from job_object_helper, return int instead (#562)

### DIFF
--- a/build/devops_gated.yml
+++ b/build/devops_gated.yml
@@ -7,7 +7,7 @@ resources:
     type: github
     name: azure/c-build-tools
     endpoint: github.com_azure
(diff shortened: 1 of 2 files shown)


anporumb (approved):
:shipit:

PR #462, opened 2026-07-21
[MrBot] Pin jemalloc 5.3.0 via c-build-tools overlay + restore c-pal priming

dcristoloveanu:
vcpkg 2026.06.01 ships **jemalloc 5.3.1**, which crashes under memory pressure (`arena_choose` returns NULL, `tcache_arena_associate` dereferences it -> AV; downstream aborts `0xc0000409`). This pins jemalloc back to **5.3.0** via the c-build-tools overlay port and restores the c-pal priming that 5.3.0's Windows/MSVC TSD lazy-init race needs.

- `deps/c-build-tools` -> master (overlay_ports/jemalloc pinning 5.3.0 [Azure/c-build-tools#437] + the `use_vcpkg` fix so the overlay actually applies [Azure/c-build-tools#438]).
- `deps/c-pal` -> master (jemalloc priming restored [Azure/c-pal#631]).
- Sync the `c_build_tools` pipeline-resource ref in `build/*.yml`.
- `deps/vcpkg` stays at 2026.06.01.

Follow-up to the merged vcpkg 2026.06.01 bump. Work item 38196464.

### DIFF
--- a/build/devops_gated.yml
+++ b/build/devops_gated.yml
@@ -22,7 +22,7 @@ resources:
     type: github
     name: azure/c-build-tools
     endpoint: github.com_azure
-    ref: cfe1933fb7f3959b798240ab29a34ad03499cf07
+    ref: a745a70ad63058276e259e8130bd36fcdc2d039e
 
 jobs:
 - template: /pipeline_templates/build_all_flavors.yml@c_build_tools


dcristoloveanu:
Abandoning: this is purely a deps bump (c-build-tools overlay + c-pal priming submodule pointers + the pipeline ref that tracks them). It will be covered by the normal dependency propagation. The overlay/priming fix rides in via c-build-tools (Azure/c-build-tools#437, #438) and c-pal (Azure/c-pal#631) once propagated.

